PR #103, opened 2017-10-04
Add bearing information to CSV export when quiver data is displayed.

clydeclements:
Users have requested to have the bearing information included in the CSV output.


htmlboss:
I also threw in a fix for the variable selector in the Area Plot

PR #142, opened 2017-11-16
Reverted landing page to "bland style".

htmlboss:
* Also, Scotty was fired :P


htmlboss:
Consultation of landing page will start in 2 weeks.

PR #146, opened 2017-12-14
Second iteration of subset feature + pan flickering bugfix

htmlboss:
* Date ranges can now be subset too.
* Fixed map pan/zoom flicker issue.
* Added new buttons to map (fullscreen, reset rotation, pan to origin).
* Minor changes to CSS for consistency.


htmlboss:
The commit containing the first subsetting iteration should be tagged as a version bump.

PR #165, opened 2018-01-17
Added colourmap dropdown to Line Window, and .csv fix.

htmlboss:
* Works for transects and hovmoller diagram


htmlboss:
We should probably merge this with master (if it's tested), just so if an issue crops up, we can revert the PR without going too far back with commits.

PR #197, opened 2018-02-07
Massive Update

htmlboss:
### Testing would be appreciated :)
* User can tweak data interpolation options.
* Refactored sidebar into a tabbed layout for reduced clutter.
* Added checks to willReceiveProps if the new props are different.
* Added babel-react-optimize-preset to package.json and babel.rc.
* Fixed broken area maps induced from colourmap selection patch.
* Minor performance pass.
* Minor css tweaks.
* General cleaning.
* Upgraded react to 16.2
* Other various package upgrades.
* node_modules folder needs to be deleted and run npm install


Jeffreydaw:
I think that the "map settings" tab and the "Data Selection" tab should be reversed. (ie so the Data Selection tab is on the far left). I think that would feel more natural since the most commonly used tabs on most applications are on the far left.

htmlboss:
Lol forgot to change that. Tyty.

htmlboss:
Fixed tab order

Jeffreydaw:
The side by side view area plots are still not working properly. They are only working if the user plots using the same dataset. If 2 different datasets are selected then it breaks (ex GIOPS daily & GIOPS Monthly). 

Ok so I have a few suggestions for little things:
* I think the compare dataset option should not be at the top of the dataset selection. Again it is less important than the actual dataset selection so I would like to see it below the "Main Map" setting. However, I think it's placement once selected is good. 
* French translation. The "map settings" tab and the "Data Selection" tab should have the French translation set (I know I'm bad for this too.) 
* "map settings" tab and the "Data Selection" tabs are a to big for smaller screens
* we need to talk about if we need to tell users about the changes to the default interpolation

Issue #45, opened 2017-07-28
map colour avarages can cross land 


Jeffreydaw:
The way the colors are displayed can cause the data to be picked up across land boundaries when they are thin. 

This can lead to problems in areas such as the Panama canal where there is hot water on one side of the land and cold water on the other. But because the only thing that is done to pick the color is to look at all the data within a certain radius and to average based on the weight of how close they are. If the land is thinner than the specified radius the data can be accessed across it leading to misleading results (ex, hot looking water on the side of the land where the water is cold). 

to fix this the colour selection tools need to know when they are next to land instead of just being covered by a mask. 

Discovered when talking to Geoff about  bug https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/7

Jeffreydaw:
This problem is still present in version [v2.1.3](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/tree/v2.1.3) and has been present since before we started using version markers (ie pre [v2.1.0](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/tree/v2.1.0)   

the image below shows an example of this problem. the pink arrow shows where the map is rendering its circle. And the green arrow shows the side of the peninsula where that data point it is using is located. 
![land jumping](https://user-images.githubusercontent.com/11149030/32326258-27b7052c-bfb5-11e7-829b-395b1b848adf.png)
![land jumping2](https://user-images.githubusercontent.com/11149030/32326611-70ac5060-bfb6-11e7-8597-b2fe0f570d3f.png)

Jeffreydaw:
@htmlboss did you fix this? It does not look like it's a problem anymore in your new code. But you make no reference to fixing it in your pull request. 

![fixedlandcrossings](https://user-images.githubusercontent.com/11149030/36543145-78ef6dc0-17bd-11e8-8e9d-51c6b9af01a6.png)

Jeffreydaw:
this change has been rolled out in version [v2.3.0](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/releases/tag/v2.3.0)

htmlboss:
Appears to be fixed now. We can reopen if needed.

Jeffreydaw:
Note this problem is probably not fixed just reduced

PR #200, opened 2018-02-14
Interpolation

htmlboss:
* User can tweak data interpolation options.
* Refactored sidebar into a tabbed layout for reduced clutter.
* Added checks to willReceiveProps if the new props are different.
* Added babel-react-optimize-preset to package.json and babel.rc.
* Fixed React warnings about missing key props.
* All components issuing Ajax calls now internally track their mounted state. This is to prevent an un-mounted component from attempting to change it's state.
* Minor performance pass.
* Minor css tweaks.
* General cleaning.
* Upgraded react to 16.2
* Other various package upgrades.
* node_modules folder needs to be deleted and run npm install


Jeffreydaw:
I've added some versioning to the API

Jeffreydaw:
this seems to have fixed or at least partly fixed bug #45 colormap can cross land.

PR #206, opened 2018-03-07
Location is plotted for Profile and CTD.

htmlboss:
* Works for multiple points.
* As normal, the location map is toggleable.
* Changed basemap of minimap to realistic bluemarble. This applies to all plots.


Jeffreydaw:
@htmlboss This is broken, see issue #223

htmlboss:
Need to see error messages. See #223 comment

PR #225, opened 2018-04-04
Cumulative update.

htmlboss:
* Shows the locations of selected points on a minimap for Profile, CTD, TS Diagram, Virtual Mooring, and Sound Speed Profile. (more otw).
* Works for multiple points.
* As normal, the location map is toggleable.
* Changed basemap of minimap to realistic bluemarble. This applies to all plots.
* Implemented Hovmoller difference plot when in compare mode.
* Rendering optimization.
* Speculative fix for fail image incorrectly showing up.
* Changed loading spinner to a more descriptive one.
* Removed text messages shown under loading/fail images in favour of a new error handling system.
* Experimental implementation of graceful error handling. This system is designed to present the user with readable messages for common errors.
* Minor edit to how subset files are downloaded. No difference to user.
* Updated landing page.
* Normalized date picker widget colour scheme to match the rest of the site.
* General code cleanup.


Jeffreydaw:
I've tested this and it looks like it works.

oceannavigator/frontend/src/components/TimePicker.jsx
```diff
@@ -18,7 +18,6 @@ import "jquery-ui/../i18n/datepicker-fr.js";
 import "jquery-ui/../i18n/datepicker-fr-CA.js";
 
 const i18n = require("../i18n.js");
-const stringify = require("fast-stable-stringify");
```
Jeffreydaw: @htmlboss I don't think this line should have been removed. I believe it is the cause of issue #227

Issue #230, opened 2018-04-12
Polar projection area plots to small 


Jeffreydaw:
when ploting an area from the polar projection the plot that gets created is allways centered at the pole.  This can lead to area plots that don't really zoom inenough to be usefull (see below) ![image](https://user-images.githubusercontent.com/11149030/38688908-d0fb61c8-3e54-11e8-8ca3-0b28b2141faf.png)

![image](https://user-images.githubusercontent.com/11149030/38688939-e0a912b4-3e54-11e8-9706-6caa9cdead5d.png)

Jeffreydaw:
fixed in commit f79466438b12780bead00e9afbf6ab106ae916ab
the new version will plot as it use to if the area is on or close to the pole (within 5 deg lat) but will plot as the rest of the world does outside of that region.

Jeffreydaw:
fixed in release [v2.3.3](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/releases/tag/v2.3.3)

PR #273, opened 2018-07-06
pole plot fix

Jeffreydaw:
this is to fix a problem related to making plots in the polar projections. 
the issue #230 had not been fixed properly.


Jeffreydaw:
reviewed by @htmlboss

PR #336, opened 2018-09-10
Weekly Update

htmlboss:
* Fixed name mismatch in subsetArea function binding.
    * Should fix the subset button doing nothing.


Jeffreydaw:
oh awesome! I'll give it a test.

PR #360, opened 2019-01-10
Adding today button to dev

NoahGallant-MUN:
Completed testing on today button... No problems encountered.
Merging with dev for next release.

This button is for the time picker and just sends the user to todays date if it exists in the dataset. (This was part of Erica's wish list)


dwayne-hart:
This code is approved.

PR #425, opened 2019-05-16
Implementing higher zoom levels for coastal datasets

### LINKED ISSUE #4: Southern Hemisphere misaligned
Major Bug: The view and the data are offset from each other in the main map view for the Southern Hemisphere. 

**Details**: 
The details of this bug are a little strange. The problem only seems to exist at the max zoom. It is most visible when a point near the equator is viewed with the Basemap of "Esri world Imagery" is selected. 

At max zoom, the view seems to be shifted about 70Km West from the data in the southern hemisphere. For the Basemaps of "Esri world Imagery" and  "Esri Ocean Basemap" the problem is more visible since the basemap is in the correct location but the data and the Bathymetry maps are shifted. for the "ETOPO1 Topography" Basemap the problem is harder to see since the basemap is shifted the same amount as the layers on top.  

The problem only exists in the global view map projection (the pole projections are fine) 

Proving that the data is shifted from the view can be done using the "line" drawing tool. 

selection_007

selection_005

In the above image the red rectangle shows where the map was displaying the land but the green rectangle shows where the land was accorading to the data. [link](http://navigator.oceansdata.ca/?query=%7B%22center%22%3A%5B%22118.9051%22%2C%22-0.6162%22%5D%2C%22zoom%22%3A8%2C%22dataset%22%3A%22giops_day%22%2C%22projection%22%3A%22EPSG%3A3857%22%2C%22time%22%3A484%2C%22variable%22%3A%22votemper%22%2C%22scale%22%3A%5B-5%2C30%5D%2C%22vectortype%22%3Anull%2C%22vectorid%22%3Anull%2C%22subquery%22%3A%7B%22selected%22%3A1%2C%22scale%22%3A%22-5%2C30%2Cauto%22%2C%22colormap%22%3A%22default%22%2C%22starttime%22%3A460%2C%22showmap%22%3Atrue%2C%22surfacevariable%22%3A%22none%22%2C%22linearthresh%22%3A200%2C%22size%22%3A%2210x7%22%2C%22dpi%22%3A72%2C%22depth_limit%22%3Afalse%7D%2C%22showModal%22%3Atrue%2C%22modal%22%3A%22line%22%2C%22names%22%3A%5B%5D%2C%22line%22%3A%5B%5B%5B-0.315750470262671%2C118.19096454986727%5D%2C%5B-0.315750470262671%2C119.17973408111727%5D%2C%5B-0.315750470262671%2C119.47636494049227%5D%2C%5B-0.31025738817055526%2C119.77848896392976%5D%2C%5B-0.30476430322667625%2C120.02568134674229%5D%2C%5B-0.31025738817055526%2C120.45964130767977%5D%5D%5D%7D)

[shift near the equator](http://navigator.oceansdata.ca/?query=%7B%22center%22%3A%5B%22118.9051%22%2C%22-0.6162%22%5D%2C%22zoom%22%3A8%2C%22dataset%22%3A%22giops_day%22%2C%22projection%22%3A%22EPSG%3A3857%22%2C%22time%22%3A484%2C%22variable%22%3A%22votemper%22%2C%22scale%22%3A%5B-5%2C30%5D%2C%22vectortype%22%3Anull%2C%22vectorid%22%3Anull%7D)
url

Problem first noticed in Verson: https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/commit/2ae745d475744bcf4770982f1de55c05b3b29238

### LINKED ISSUE #356: Implement vector-based bathymetry contours
## Preamble

The idea is that instead of generating bathymetry tiles for every zoom level, we create a single GeoJSON file with the bathymetry contours for the globe and tell OpenLayers to render it as a vector image. Free infinite zooming, which is a huge step forward in implementing coastal datasets.

I've taken a first crack at this and hopefully resolved the most difficult parts. The raw GeoJson (before I implemented a Gaussian filter) is about 450MB. This version will be stored on [Dory](http://navigator.oceansdata.ca/cdn) if anyone wants to use it in the future for super high-res stuff. It requires about 15GB RAM to generate.

world_bathymetry_contours_simplified.zip contains the simplified output of the bathymetry contours ready to be used in the Navigator. Zipped file is~5MB while the unzipped version is ~17MB (Gaussian filter did it's magic as expected).

* I've already computed the optimal output, so you probably don't need to repeat these steps. To view the output GeoJSON files, and simplify the shapes for the final output: [https://mapshaper.org/](https://mapshaper.org/). Make sure you select "visvalingam/weighted area".

## What you need to do
* Write an API extension to send the unzipped file I have attached here (MAKE SURE YOU ENABLE ZIP COMPRESSION IN THE HTTP HEADERS).
* Get OpenLayers to send an API query to get the file, and save it to the user's long-term browser cache (something about local data storage).
* Replace the current bathymetry tile layer with a new vector-based layers (look at OpenLayers docs for this).
* Also implement the loading screen for the website so we can load resources and ensure the user knows what's going on.

## Misc. documentation
* [scipy gaussian filter](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html)

* Python 3 script to do the conversion:
```python
#!/usr/bin/env python

import sys
import geojson as gj
import xarray as xr
from scipy.ndimage.filters import gaussian_filter

if len(sys.argv) != 2:
	print("Usage: %s Gaussian filter sigma (5.0 is recommended)")
	exit()

sigma = float(sys.argv[1])
if sigma < 0.0:
	print("Gaussian filter sigma must be >= 0.0")
	exit()

with xr.open_dataset('/data/hdd/misc/ETOPO1_Bed_g_gmt4.grd') as ds:
	
	print("Filtering...")
	ds['z'][:] = gaussian_filter(ds['z'], sigma)
	ds['z'] *= -1
	data = ds['z'][::-1, :]

	print("Calculating contours...")
	c = xr.plot.contour(data, x='x', y='y', levels=[100, 200, 500, 1000, 2000, 3000, 4000, 5000, 6000])

	print("Collecting line geometry...")
	line_features = []
	for collection in c.collections:
		paths = collection.get_paths()

		for path in paths:
			v = path.vertices
			coords = []
			
			for i in range(len(v)):
				lat = v[i][0]
				lon = v[i][1]
				coords.append((lat, lon))

			line = gj.LineString(coords)
			line_features.append(gj.Feature(geometry=line))


	feature_collection = gj.FeatureCollection(line_features)

	print("Dumping to file...")
	geojson_dump = gj.dumps(feature_collection, sort_keys=True)
	with open('out.geojson', 'w') as out:
		out.write(geojson_dump)

	print("Done!")

```

[world_bathymetry_contours_simplified.zip](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/files/2699225/world_bathymetry_contours_simplified.zip)

### LINKED ISSUE #357: Implement a loading screen.
Since we are moving to a vector-based approach for OpenLayers, there is some pre-requisite data the user's browser will need to download before the map is ready (Hopefully a max of 15MB placed into local browser storage for long-term). A loading screen is ideal for this. Something like a plain page with a logo and a loading bar/circle will suffice. It should then fade into the main map page. This should only happen the first time the user enters the site (since no data has been stored yet), or after they clear their browser cache. Other times there shouldn't be any waiting to get to the map.

### LINKED ISSUE #358: Implement vector-based topography
## Converted output:
[out_topo.geojson.gz](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/files/2703005/out_topo.geojson.gz)

## Conversion script
```python
#!/usr/bin/env python

import sys
import geojson as gj
import xarray as xr
from scipy.ndimage.filters import gaussian_filter

with xr.open_dataset('/data/hdd/misc/ETOPO1_Bed_g_gmt4.grd') as ds:
	
	print("Filtering...")
	ds['z'][:] = gaussian_filter(ds['z'], 0.25)

	print("Calculating contours...")
	c = xr.plot.contour(ds['z'], x='x', y='y', levels=[0])

	print("Collecting line geometry...")
	line_features = []
	for collection in c.collections:
		paths = collection.get_paths()

		for path in paths:
			v = path.vertices
			coords = []
			
			for i in range(len(v)):
				lat = v[i][0]
				lon = v[i][1]
				coords.append((lat, lon))

			line = gj.LineString(coords)
			line_features.append(gj.Feature(geometry=line))


	feature_collection = gj.FeatureCollection(line_features)

	print("Dumping to file...")
	geojson_dump = gj.dumps(feature_collection, sort_keys=True)
	with open('out_topo.geojson', 'w') as out:
		out.write(geojson_dump)

	print("Done!")

```

jamesprayner:
## What is this PR doing?

Resolves #356
Resolves #357 
Resolves #358 

Implementing a way to display zoom levels past level 8 (currently levels up to 13 are working with this update). Next step is to add the coastal data to dory so this feature can be used efficiently.

This also resolves #4 (finally!) by changing the source of the tiles on zoom level 8.

## Why have you chosen this approach?
The original idea in #356 was to use a vector layer with a single geojson file to display bathymetry and do something similar for #358. Turns out rendering a massive geojson file is very tough on performance, even with a simplified file. 
VectorTile layers are easier on the client and don't have a single 500MB file associated with it that needs to be downloaded before anything can be displayed. The mbt format (mapbox tiles) is a compressed version of a geojson, so they're smaller and require even less resources. 
The mbt data is stored in SQLite files with the extension ".mbtiles", if not for this each installation of the navigator would need to downloading and extracting an archive with 4^13 sub directories (this takes far too much time).

## Should reviewers focus on anything specific?
Check the sanity of the API changes, since some old code (only slightly) was changed to this working.

* [x] I have tested my changes.
* [x] (If applicable): I ran `npm run lint` and resolved all errors.


jamesprayner:
The new MBTiles files are being hosted on Dory and have been included in the install script.

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -206,9 +206,48 @@ export default class Map extends React.PureComponent {
...
+          stroke: new ol.style.Stroke({
+            color: 'rgba(0, 0, 0, 1)'
+          }),
+				  fill: new ol.style.Fill({
+					  color: 'white'
+            })
+          }),
+        source: new ol.source.VectorTile({
+          format: new ol.format.MVT(),
+          tileGrid: new ol.tilegrid.createXYZ({tileSize:512, maxZoom: 13}),
```
htmlboss: Is `tileSize` the size in pixels of the rendered tile?

routes/routes_impl.py
```diff
@@ -615,6 +622,45 @@ def bathymetry_impl(projection: str, zoom: int, x: int, y: int):
...
+  # Send blank tile if conditions aren't met
+  if (zoom < 7) or (projection != "EPSG:3857"):
+    return send_file(shape_file_dir + "/blank.mbt")
+
+# Send file if cached or select data in SQLite file
+  if os.path.isfile(directory):
+    return send_file(directory)
+  else:
+    y = (2**zoom-1) - y
+    connection = sqlite3.connect(shape_file_dir + "/{}.mbtiles".format(tiletype))
```
htmlboss: Is it possible `.connect` may throw an exception? Should we check for a failed connection?

routes/routes_impl.py
```diff
@@ -589,6 +592,8 @@ def topo_impl(projection: str, zoom: int, x: int, y: int, shaded_relief: bool):
     """
         Generates topographical tiles
     """
+    if zoom > 7:
+        return send_file("/opt/tiles/blank.png")
```
htmlboss: Is this file on dory?

routes/routes_impl.py
```diff
@@ -615,6 +622,45 @@ def bathymetry_impl(projection: str, zoom: int, x: int, y: int):
...
+    connection = sqlite3.connect(shape_file_dir + "/{}.mbtiles".format(tiletype))
+    selector = connection.cursor()
+    sqlite = "SELECT tile_data FROM tiles WHERE zoom_level = {} AND tile_column = {} AND tile_row = {}".format(zoom, x, y)
+    selector.execute(sqlite)
+    tile = selector.fetchone()
+    if tile == None:
+        return send_file(shape_file_dir + "/blank.mbt")
+
+    # Write tile to cache and send file
+    if os.path.isdir(basedir):
```
htmlboss: Similar comment as earlier, we can simplify this like so:

```python
if not os.path.isdir(basedir):
  os.makedirs(basedir)
```

routes/routes_impl.py
```diff
@@ -615,6 +622,45 @@ def bathymetry_impl(projection: str, zoom: int, x: int, y: int):
...
+      pass
+    else:
+      os.makedirs(basedir)
+    with open(directory + ".pbf", 'wb') as f:
+      f.write(tile[0])
+    with gzip.open(directory + ".pbf", 'rb') as gzipped:
+      with open(directory, 'wb') as tileout:
+        shutil.copyfileobj(gzipped, tileout)
```
htmlboss: If I'm reading this code correctly, we're:
* Writing out the fetched tile from sqlite as a `.pbf`
* Reading it back in through gzip.
* Copying the contents of that `.pbf` into the requested tile file.

Do I have this correct?

routes/routes_impl.py
```diff
@@ -615,6 +622,45 @@ def bathymetry_impl(projection: str, zoom: int, x: int, y: int):
         img = plotting.tile.bathymetry(projection, x, y, zoom, {})
         return _cache_and_send_img(img, f)
 
+def mbt_impl(projection: str, tiletype: str, zoom: int, x: int, y: int):
+  """
+       Serves mbt files
+  """
+  cache_dir = current_app.config['CACHE_DIR']
+  shape_file_dir = current_app.config['SHAPE_FILE_DIR']
+  directory = str(os.path.join(cache_dir, request.path[1:]))
```
htmlboss: Is `directory` actually a directory path, or does it end with some file extension? I'm asking because on line 660 and 662 `directory` is being used as a filepath.

oceannavigator/frontend/src/components/Map.jsx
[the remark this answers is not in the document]
jamesprayner: Yes. This is the correct size for the mbtiles files that were created.

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -934,6 +975,27 @@ export default class Map extends React.PureComponent {
         })
       );
 
+      // Update Hi-res bath layer
+      this.layer_bathshapes.setSource(
+        new ol.source.VectorTile({
+          format: new ol.format.MVT(),
+          tileGrid: new ol.tilegrid.createXYZ({tileSize:512, maxZoom: 14}),
```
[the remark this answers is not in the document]
jamesprayner: Yes, they should be the same `maxZoom` as the projection. I've defined `this.vectorTileGrid` outside  and called it anytime its needed

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -934,6 +975,27 @@ export default class Map extends React.PureComponent {
...
+          tilePixelRatio: 8,
+          url: `/api/v1.0/mbt/${this.props.state.projection}/bath/{z}/{x}/{y}`,
+        })
+      );
+
+      // Update Hi-res land layer
+      this.layer_landshapes.setSource(
+        new ol.source.VectorTile({
+          format: new ol.format.MVT(),
+          tileGrid: new ol.tilegrid.createXYZ({tileSize:512, maxZoom: 14}),
```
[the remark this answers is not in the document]
jamesprayner: See previous.

oceannavigator/oceannavigator.cfg
```diff
@@ -11,5 +11,4 @@ CLASS4_CATALOG_URL = "http://trinity:8080/thredds/catalog/class4/catalog.xml"
 CLASS4_URL = "http://trinity:8080/thredds/dodsC/class4/%s.nc"
 OBSERVATION_AGG_URL = "http://trinity:8080/thredds/dodsC/misc/observations/aggregated.ncml"
 ETOPO_FILE = "/data/hdd/misc/etopo_%s_z%d.nc"
-
-
+SHAPE_FILE_DIR = "/data/hdd/misc/shapes"
```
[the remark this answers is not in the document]
jamesprayner: It didn't before you mentioned it, I just added it.
It's worth noting, however, that the directory `/data/hdd/` doesn't exist on dory, so the config file will be slightly different. The shape file directory is located in `/data/misc/shapes` on dory.

routes/routes_impl.py
```diff
@@ -615,6 +622,45 @@ def bathymetry_impl(projection: str, zoom: int, x: int, y: int):
...
+  # This is a funky way of splitting the string to only include characters up to the 10th instance of '/'
+  basedir = "/".join(directory.split("/", 10)[:10])
+
+  # Send blank tile if conditions aren't met
+  if (zoom < 7) or (projection != "EPSG:3857"):
```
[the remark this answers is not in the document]
jamesprayner: Not initially. The way that `mbtiles` supports other projections is sort of funky, but should be a goal in the future.

routes/routes_impl.py
```diff
@@ -615,6 +622,45 @@ def bathymetry_impl(projection: str, zoom: int, x: int, y: int):
...
 
+def mbt_impl(projection: str, tiletype: str, zoom: int, x: int, y: int):
+  """
+       Serves mbt files
+  """
+  cache_dir = current_app.config['CACHE_DIR']
+  shape_file_dir = current_app.config['SHAPE_FILE_DIR']
+  directory = str(os.path.join(cache_dir, request.path[1:]))
+  # This is a funky way of splitting the string to only include characters up to the 10th instance of '/'
+  basedir = "/".join(directory.split("/", 10)[:10])
```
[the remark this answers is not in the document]
jamesprayner: The 10th slash was the last one for the directory path without the filename. I realised this is a terrible way to do this, so I replaced it with:
`basedir = requestf.rsplit("/", 1)[0]`
which splits at the last instance of `/` and gives the first half.

routes/routes_impl.py
jamesprayner: No, this was a mistake. This file is included in the shapes tar file on the installation script. I'll change the string to include the global variable `shape_file_dir`

routes/routes_impl.py
jamesprayner: I don't think there's anyway outside of having to check if the mbtiles files exist before each mbt file can be sent. There's no such check for old topo files (they're just plonked in a `with` block), so I didn't see any reason to add one here. The only way a failed connection could occur is just that the files don't exist in the place they should, which means the navigator isn't installed correctly.

routes/routes_impl.py
jamesprayner: Yes, that's what's happening. If you have a nicer way off the top of your head I'm all ears

routes/routes_impl.py
jamesprayner: Sorry about this. I've changed some of the variable names just now so it makes a bit more sense:
`requestf` is the mbt file (no extension) that is being requested by the user
`basedir` is the directory in which that file is stored, or where it is meant to be stored

plotting/tile.py
```diff
@@ -218,13 +218,16 @@ def plot(projection, x, y, z, args):
     data = data.transpose()
     xpx = x * 256
     ypx = y * 256
+    
+    if z > 7:
```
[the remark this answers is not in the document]
jamesprayner: Okay, noted. Changed this and the next mention of this.

routes/routes_impl.py
jamesprayner: Changed this and the previous mention of this.

routes/routes_impl.py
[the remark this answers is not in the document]
htmlboss: Ah I see. No problem since the coastal datasets coming in don't cover those areas.

routes/routes_impl.py
htmlboss: I'm wondering can we just write directly to the tile file and skip the pbf entirely?

htmlboss: Seems like gzip doesn't like bytesio objects for some reason. I'll create an issue and we can look at this later.

PR #429, opened 2019-05-23
Update datasetconfig.json

htmlboss:
* Added new calculated variables to supported datasets.
* Switch from navigator domain to localhost (much better performance).


htmlboss:
I've tested this on my local machine, and everything is working as intended. I'll go ahead and override the review requirement.

PR #434, opened 2019-05-27
Lat / Lon variable fix for subsetting 

jamesprayner:
- Something with Geoff's code changes the list `variable_list` to not include the lat lon variables (I've no idea).

- This PR makes the subset function search through the superset of variable_list for the lat lon variables.

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -128,8 +128,8 @@ class NetCDFData(Data):
         variable_list = [v.key for v in self.variables]
 
         # Get lat/lon variable names from dataset (since they all differ >.>)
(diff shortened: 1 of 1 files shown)


data/netcdf_data.py
```diff
@@ -128,8 +128,8 @@ def find_variable(substring: str, variables: list):
         variable_list = [v.key for v in self.variables]
```
htmlboss: Is `variable_list` being used elsewhere below?

jamesprayner: Doesn't look like it, should I comment it out or remove it?

PR #436, opened 2019-05-28
netcdftime is now cftime

jmunroe:
In late 2016, cftime was split out of the netcdf4-python package (https://github.com/Unidata/cftime).  This PR makes that change allowing us to use more recent versions of the netcdf4 package.


jmunroe:
Currently, do we have netcdf4=1.3.1 in the conda repo?  This PR will be needed when and if we go to netcdf4>=1.4

Do we have automation tools for bundling conda environments into a downloadable asset? (I am not keen on tarballs/zip files for development environments but I can see the value to deploy the application)

utils/misc.py
```diff
@@ -2,9 +2,10 @@
 import os
 import time
 from flask import current_app
-from pykml import parser
+# from pykml import parser
```
htmlboss: @jamesprayner will have a pr up to fix this.

Issue #463, opened 2019-07-11
API R scripts not working


htmlboss:
All URLs generated by the R API Scripts you can request from the navigator currently return a 502 gateway error.

NoahGallant-MUN:
#487

PR #464, opened 2019-07-11
Band-aid fix to get class4 script gen working

jamesprayner:
Enables class4 script generation. R Scripts are still broken (#463).

### DIFF
--- a/oceannavigator/frontend/src/components/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map.jsx
@@ -587,7 +587,8 @@ export default class Map extends React.PureComponent {
         if (feature.get("type") != null) {
           switch(feature.get("type")) {
             case "class4":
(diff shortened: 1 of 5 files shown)


oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -587,7 +587,8 @@ export default class Map extends React.PureComponent {
         if (feature.get("type") != null) {
           switch(feature.get("type")) {
             case "class4":
-              content.push(feature.get("id"));
+              const class4id = feature.get("id").replace("/", "_");
```
htmlboss: I think a brief comment in the code explaining why we're doing the replacement would be good for the future.

htmlboss:
Just emphasising this is a temporary fix until @jamesprayner has the script gen code re-written.

Issue #268, opened 2018-06-19
Stats Table broken for Area Plot


NoahGallant-MUN:
Broken for all variables
gives the same value for all statistics

may be related to issue #236

NoahGallant-MUN:
The stats table won't even load anymore

htmlboss:
We've decided to re-write the stats file completely. I'll be doing up a design doc for the public api. For the first implementation, we'll most likely revert to similar behaviour before #236 was fixed. It'll be modular enough to add dateline support if users need it. Performance-wise the stats api is doing very easy calculations (min, max, mean, median, etc), so results should be presented near instantly.

PR #478, opened 2019-08-28
Rewrite stats+unit tests

halehkh65:
Stats table works and the statistical calculations for the requested area of each dataset are shown in the stats table.


jmunroe:
I'll take a look at these changes.

jmunroe:
This has gone stale and is now out of sync with rest of the project. This is issue is being tracked as #268 and will be resubmitted again.

PR #510, opened 2019-10-11
Sound Channel Calculations

NoahGallant-MUN:
This pull request adds additional capabilities to the calculation layer of the navigator.
Most notably, the addition of the tilda (~) operator to the parser. A ~ placed in front of a function argument variable (a variable actually in the dataset), will provide the calculation function with all depths of that particular variable, rather than just the default horizontal slice.
The addition of the ~ operator, rather than taking other approaches, was largely because the issue at hand was the lack of 3D data for variables, the tilda solves this by allowing the parser to treat it differently than a typical argument.

Using the ~ operator, 4 different Sound Speed related calculation functions were written allowing for the display of the: Sound Channel Axis, Sonic Layer Depth, Critical Depth, and Depth Excess were added.

The only problem this pull request may pose on the navigator, is that the speed of the calculation decreases with each variable. For example, the Depth Excess requires calculating first the Sound Channel Axis, followed by the sonic layer depth, and then the critical depth. This results in a relatively slow (but not unbearable) speed.
Methods of reducing this performance hit were investigated, but those implemented showed equal or reduced performance.
However, despite this, additional performance improvement measures should be attempted.

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -88,16 +88,16 @@ def sspeedmin(depth, lat, lon, temperature, salinity):
     
     for x in range(speed.shape[0]):
         for y in range(speed.shape[1]):
-            min_val = np.nanmin(speed[x][y])
-            idx = np.where(speed[x][y] == min_val)
+            min_val = np.nanmin(speed[x,y])
+            idx = np.where(speed[x,y] == min_val)
             if (np.isnan(min_val)):
-                speed[x][y] = np.nan
+                speed[x,y] = np.nan
             elif idx[0].shape[0] > 1:
                 idx = idx[0][0]
 
-                speed[x][y] = depth.values[idx]
+                speed[x,y] = depth.values[idx]
             else:
-                speed[x][y] = depth.values[idx]
+                speed[x,y] = depth.values[idx]
             #speed[x][y] = depth[np.where(speed[x][y] == np.nanmin(speed[x][y]))]  #np.nanmin(speed[x][y])
     
     speed = speed.transpose()
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -125,7 +125,8 @@ def soniclayerdepth(depth, lat, temperature, salinity):
     #for x in range(speed.shape[0]):
     #    for y in range(speed.shape[1]):
     # TESTING
-    sld_array = np.nan((speed.shape[-2], speed.shape[-1]))
+    #sld_array = np.nan((speed.shape[-2], speed.shape[-1]))
+    sld = np.nan
     for x in range(speed.shape[-1]):
         for y in range(speed.shape[-2]):
             sca_value = np.nanmin(speed[x,y])
@@ -144,7 +145,7 @@ def soniclayerdepth(depth, lat, temperature, salinity):
                 else:
                     sld_idx = np.where(subset == sld_value)[0][0]
                     sld = depth.values[sld_idx]
-                    sld_array[x,y] = sld
+                    speed[:, y, x] = sld
 
     #speed = speed.transpose()
     #speed = speed[0]
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -264,6 +264,8 @@ def sscp(depth, lat, temperature,salinity):
     temperature: The temperatures(s) (at all depths) in celsius
     salinity: The salinity (at all depths) (unitless)
     """
+
+    speed = sspeed(depth, lat, temperature, salinity)
     max_depth = np.abs(depth - 1000).argmin()
     temp_subset = speed[:max_depth]
     salinity_subset = salinity[:max_depth]
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -229,33 +229,15 @@ def sspeedmin(depth, lat, temperature, salinity):
     """
     
     speed = sspeed(depth, lat, temperature, salinity)
-    #speed = speed.transpose()
-    
-    #for x in range(speed.shape[0]):
-    #    for y in range(speed.shape[1]):
     for x in range(speed.shape[-1]):
         for y in range(speed.shape[-2]):
-            #min_val = np.nanmin(speed[:,y,x])
-            #idx = np.where(speed[:,y,x] == min_val)
-            
-            #if (np.isnan(min_val)):
-            #    speed[:,y,x] = np.nan
-            #elif idx[0].shape[0] > 1:
-            #    idx = idx[0][0]
-            #    speed[:,y,x] = depth.values[idx]
-            # Find sca idx
             idx = find_sca_idx(speed[:,y,x])
             if np.isnan(idx):
                 speed[:,y,x] = idx
             else:
                 speed[:,y,x] = depth.values[idx]
-            #speed[x][y] = depth[np.where(speed[x][y] == np.nanmin(speed[x][y]))]  #np.nanmin(speed[x][y])
     
-    #speed = speed.transpose()
-
-    speed = speed[0]
-
-    return speed
+    return speed[0]
 
 def soniclayerdepth(depth, lat, temperature, salinity):
     """
@@ -278,21 +260,24 @@ def soniclayerdepth(depth, lat, temperature, salinity):
     sld = np.nan
     for x in range(speed.shape[-1]):
         for y in range(speed.shape[-2]):
-            sca_value = np.nanmin(speed[:,y,x])
-            sca_idx = np.where(speed[:,y,x] == sca_value)
+            #sca_value = np.nanmin(speed[:,y,x])
+            #sca_idx = np.where(speed[:,y,x] == sca_value)
             
-            if (np.isnan(sca_value)):
-                pass
-            else:
-                sca_idx = sca_idx[0][0]
+            sca_idx = find_sca_idx(speed[:,y,x])
 
-                subset = speed[:,y,x][0:int(sca_idx) + 1]
-                sld_value = subset.max()
-                
-                if (np.isnan(sld_value)):
-                    pass
+            if (np.isnan(sca_idx)):
+                speed[:,y,x] = sca_value
+            else:
+                #sca_idx = sca_idx[0][0]
+
+                #subset = speed[:,y,x][0:int(sca_idx) + 1]
+                #sld_value = subset.max()
+                sld_idx = find_sld_idx(sca_idx, speed[:,y,x])
+
+                if (np.isnan(sld_idx)):
+                    speed[:,y,x] = sld_idx
                 else:
-                    sld_idx = np.where(subset == sld_value)[0][0]
+                    #sld_idx = np.where(subset == sld_value)[0][0]
                     sld = depth.values[sld_idx]
                     speed[:, y, x] = sld
 
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -226,6 +226,54 @@ def cd_interpolation(cd_idx, sld_idx, speed_point, depth):
 
     return cd_depth
 
+
+def __is_max(p1, p2, p3):
+    if p2 > p1 and p2 > p3:
+        return 1
+    else:
+        return np.nan
+
+def sscp_point(sspeed):
+    """
+    sscp - Secondary Sound Channel Potential (above 1000m)
+    This function determines if a secondary sound channel could exist in a water column
+
+    Parameters:
+    sspeed: np.array of sound speeds for a single water column
+    
+    Ensures:
+    Returns either np.nan or 1
+    """
+
+    for d in range(sspeed.shape[0]):
+        if d is not 1 and d is not sspeed.shape[0]:
+
+            print(something)
+            return __is_max(sspeed[d-1], sspeed[d], sspeed[d+1])
+        else:
+            return np.nan
+
+def sscp(depth, lat, temperature,salinity):
+    """
+    Determines if a Secondary Sound Channel could exist (above 1000m)
+
+    Parameters:
+    depth: The depth(s) in meters
+    lat: The latitude(s) in degrees North
+    temperature: The temperatures(s) (at all depths) in celsius
+    salinity: The salinity (at all depths) (unitless)
+    """
+    max_depth = np.abs(depth - 1000).argmin()
+    temp_subset = speed[:max_depth]
+    salinity_subset = salinity[:max_depth]
+    speed = sspeed(depth, lat, temp_subset, salinity_subset)
+
+    for x in range(speed.shape[-1]):
+        for y in range(speed.shape[-2]):
+            if (speed[:,y,x].size - np.count_nonzero(np.isnan(speed[:,y,x]))) != 0:
+                speed[:,y,x] = sscp(speed[:,y,x])
+
+                
 def soundchannelaxis(depth, lat, temperature, salinity):
     """
     Finds the global minimum of the speed of sound
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -74,7 +74,8 @@ def sspeed(depth, latitude, temperature, salinity):
     return np.array(speed)
 
 
-
+def count_numerical_vals(array):
+    return array.size - np.count_nonzero(np.isnan(array))
 
 def find_sca_idx(speed):
     """
@@ -144,7 +145,7 @@ def find_cd_idx(sca_idx, sld_idx, speed):
     sld_value = speed[sld_idx]
 
     # Find total_idx
-    total_idx = speed.size - np.count_nonzero(np.isnan(speed)) - 1
+    total_idx = count_numerical_vals(speed) - 1  #speed.size - np.count_nonzero(np.isnan(speed)) - 1
     
     # Create Layer Subset
     lower_subset = speed[sca_idx + 1: total_idx + 1]
@@ -164,7 +165,7 @@ def find_cd_idx(sca_idx, sld_idx, speed):
     idx = np.abs(lower_subset - sld_value).argmin()
 
     # Find size of lower subset
-    subset_size = lower_subset.size - np.count_nonzero(np.isnan(lower_subset)) - 1
+    subset_size = count_numerical_vals(lower_subset) - 1 #lower_subset.size - np.count_nonzero(np.isnan(lower_subset)) - 1
 
     # Shift to get index for non subset
     cd_idx = total_idx - (subset_size - idx)
@@ -239,7 +240,8 @@ def soundchannelaxis(depth, lat, temperature, salinity):
     speed = sspeed(depth, lat, temperature, salinity)
     for x in range(speed.shape[-1]):
         for y in range(speed.shape[-2]):
-            if (speed[:,y,x].size - np.count_nonzero(np.isnan(speed[:,y,x]))) != 0:
+            if count_numerical_vals(speed[:,y,x]) != 0:
+            #if (speed[:,y,x].size - np.count_nonzero(np.isnan(speed[:,y,x]))) != 0:
                 idx = find_sca_idx(speed[:,y,x])
                 if np.isnan(idx):
                     speed[:,y,x] = idx
@@ -267,27 +269,29 @@ def soniclayerdepth(depth, lat, temperature, salinity):
     
     # Find speed of sound
     speed = sspeed(depth, lat, temperature, salinity)
-    
+    result = numpy.empty(speed.shape[-1], speed.shape[-2])
     for x in range(speed.shape[-1]):
         for y in range(speed.shape[-2]):
-            if (speed[:,y,x].size - np.count_nonzero(np.isnan(speed[:,y,x]))) != 0:
+            speed_point = speed[:,y,x]
+            if (count_numerical_vals(speed_point) != 0):
+            #if (speed[:,y,x].size - np.count_nonzero(np.isnan(speed[:,y,x]))) != 0:
             
-                sca_idx = find_sca_idx(speed[:,y,x])
+                sca_idx = find_sca_idx(speed_point)
 
                 if (np.isnan(sca_idx)):
-                    speed[:,y,x] = sca_idx
+                    result[x,y] = sca_idx
                 else:
-                    sld_idx = find_sld_idx(sca_idx, speed[:,y,x])
+                    sld_idx = find_sld_idx(sca_idx, speed_point)
 
                     if (np.isnan(sld_idx)):
-                        speed[:,y,x] = sld_idx
+                        result[x,y] = sld_idx
                     else:
                         sld = depth.values[sld_idx]
-                        speed[:, y, x] = sld
+                        result[x,y] = sld
             else:
-                speed[:,y,x] = np.nan
+                result[x,y] = np.nan
 
-    return speed[0] # Only return one horizontal slice
+    return result # Only return one horizontal slice
 
 def criticaldepth(depth, lat, temperature, salinity):
     """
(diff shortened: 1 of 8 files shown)


htmlboss:
> There are so many commits because they are all tiny, and I need to commit to test things in my container on dory.

No problem, we'll do a squash-and-merge into master so all these commits will be condensed into 1.

htmlboss:
Could you also update the original PR comment to include what changes, any particular design choices, and possible areas of issue in the code. It'll help to have a higher-level view of what's going on.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+                speed[x,y] = np.nan
+            elif idx[0].shape[0] > 1:
+                idx = idx[0][0]
+
+                speed[x,y] = depth.values[idx]
+            else:
+                speed[x,y] = depth.values[idx]
+            #speed[x][y] = depth[np.where(speed[x][y] == np.nanmin(speed[x][y]))]  #np.nanmin(speed[x][y])
+    
+    speed = speed.transpose()
```
htmlboss: I'm wondering why do we need to perform a transpose twice?

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+                sld_value = subset.max()
+                
+                if (np.isnan(sld_value)):
+                    pass
+                else:
+                    sld_idx = np.where(subset == sld_value)[0][0]
+                    sld = depth.values[sld_idx]
+                    speed[x,y] = sld
+
+    speed = speed.transpose()
```
htmlboss: Same comment as above for double transpose.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+    speed = speed[0]
+    return np.array(speed)
+
+
+def find_sca_idx(speed):
```
htmlboss: I think it would help future users to include what `sca` means as part of the docstring.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+                            cd_depth = np.nan
+                    else:
+                        cd_depth = np.nan
+            else:
+                cd_depth = np.nan
+
+            speed[x,y] = cd_depth
+                
+
+    speed = speed.transpose()
```
htmlboss: Same as above for the double transpose.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+                            depth_excess = np.nan
+                    else:
+                        depth_excess = np.nan
+            else:
+                depth_excess = np.nan
+
+            speed[x,y] = depth_excess
+                
+
+    speed = speed.transpose()
```
htmlboss: Same as above for double transpose.

data/calculated_parser/lexer.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, **kwargs):
                 'TIMES',
                 'DIVIDE',
                 'POWER',
+                'TILDA',
```
htmlboss: I think the correct spelling here is "tilde" (yeah English, as usual haha).

data/calculated_parser/lexer.py
```diff
@@ -30,6 +31,7 @@ def __init__(self, **kwargs):
         self.t_TIMES = '\\*'
         self.t_DIVIDE = '/'
         self.t_POWER = '\\^'
+        self.t_TILDA = '~'
```
htmlboss: Same as above for spelling correction.

data/calculated_parser/parser.py
```diff
@@ -85,6 +110,17 @@ def p_expression_variable(self, t):
                     self.data.variables[t[1]]
                     )
                 ]
+        
+
+    def p_expression_variable_all(self, t):
+        'expression : TILDA ID'
```
htmlboss: "TILDE" spelling.

data/nemo.py
```diff
@@ -236,19 +236,17 @@ def get_point(self, latitude, longitude, depth, timestamp, variable,
...
         
-
         if depth == 'bottom':
             
             if hasattr(time, "__len__"):
                 d = var[time[0], :, miny:maxy, minx:maxx]
             else:
                 d = var[time, :, miny:maxy, minx:maxx]
-
+            print(somethingafter)
```
htmlboss: A wild print statement appeared.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+
+    speed = speed[0]
+
+    return np.array(speed)
+
+def soniclayerdepth(depth, lat, lon, temperature, salinity):
+    """
+    Finds the local maxima of the speed of sound
+
+    Parameters:
```
htmlboss: Please ensure the parameter comments reflect the function arguments.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+                speed[x,y] = depth.values[idx]
+            #speed[x][y] = depth[np.where(speed[x][y] == np.nanmin(speed[x][y]))]  #np.nanmin(speed[x][y])
+    
+    speed = speed.transpose()
+
+    speed = speed[0]
+
+    return np.array(speed)
+
+def soniclayerdepth(depth, lat, lon, temperature, salinity):
```
htmlboss: Is the `lon` argument being used in this function?

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
     speed = seawater.svel(salinity, temperature, press)
     return np.array(speed)
 
 
+def sspeedmin(depth, lat, lon, temperature, salinity):
+    """
+    Finds the global minimum of the speed of sound
+
+    Parameters:
+    sspeed: Speed of Sound
```
htmlboss: Please ensure the parameter comments reflect the function arguments.

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
         press = [seawater.pres(d, latitude) for d in depth]
     except TypeError:
         press = seawater.pres(depth, latitude)
-
+    
     speed = seawater.svel(salinity, temperature, press)
     return np.array(speed)
 
 
+def sspeedmin(depth, lat, lon, temperature, salinity):
```
htmlboss: Is the `lon` argument being used in this function?

plotting/sound.py
```diff
@@ -65,16 +65,91 @@ def plot(self):
...
             np.amin(self.sspeed) - (maxspeed - minspeed) * 0.1,
             np.amax(self.sspeed) + (maxspeed - minspeed) * 0.1,
         ])
+
+        if self.query.get('annotate'):
+            # Sound Speed Minima
+            minspeed = float("{0:.2f}".format(minspeed))
+            plt.text(minspeed + 2, minpos, "Sound Channel Axis: " + str(minspeed) + 'm')
+            ax.scatter(minspeed, minpos, s=200, marker='_')
+            # ~~~~~~~~~~~~~~~~~
```
htmlboss: Would adding an empty line be sufficient in order to add white space?

plotting/sound.py
```diff
@@ -65,16 +65,91 @@ def plot(self):
...
+
+        if self.query.get('annotate'):
+            # SOUND SPEED MINIMA
+            
+            minpos = np.where(self.sspeed[0] == minspeed)
+
+            # SONIC LAYER DEPTH
+
+            # Calculate
+            soniclayerdepth_value = self.sspeed[0][0:int(minpos[0])].max()
```
htmlboss: I'm wondering what's the conceptual difference between `soniclayerdepth_value` and `soniclayerdepth`?

plotting/sound.py
```diff
@@ -65,16 +65,91 @@ def plot(self):
...
+            soniclayerdepth_idx = np.where(self.sspeed[0] == soniclayerdepth_value)
+            soniclayerdepth = self.temperature_depths.data[0][soniclayerdepth_idx][0]
```
htmlboss: Could this be wrapped into multiple functions? Maybe something like `self.__calc_soniclayerdepth_value` and `self.__calc_soniclayerdepth`?

NoahGallant-MUN:
This pull request adds additional capabilities to the calculation layer of the navigator.
Most notably, the addition of the tilda (~) operator to the parser. A ~ placed in front of a function argument variable (a variable actually in the dataset), will provide the calculation function with all depths of that particular variable, rather than just the default horizontal slice.
The addition of the ~ operator, rather than taking other approaches, was largely because the issue at hand was the lack of 3D data for variables, the tilda solves this by allowing the parser to treat it differently than a typical argument.

Using the ~ operator, 4 different Sound Speed related calculation functions were written allowing for the display of the: Sound Channel Axis, Sonic Layer Depth, Critical Depth, and Depth Excess were added.

The only problem this pull request may pose on the navigator, is that the speed of the calculation decreases with each variable. For example, the Depth Excess requires calculating first the Sound Channel Axis, followed by the sonic layer depth, and then the critical depth. This results in a relatively slow (but not unbearable) speed.
Methods of reducing this performance hit were investigated, but those implemented showed equal or reduced performance.
However, despite this, additional performance improvement measures should be attempted.

plotting/sound.py
NoahGallant-MUN: soniclayerdepth is the depth of the local maxima, whereas soniclayerdepth_value is the sound speed at the soniclayerdepth

data/calculated_parser/functions.py
NoahGallant-MUN: This was part of my initial attempts at making it work. I'm not sure what the transpose does at all... I'll investigate on monday to see if it's needed

data/calculated_parser/functions.py
NoahGallant-MUN: This was part of my initial attempts at making it work. I'm not sure what the transpose does at all... I'll investigate on monday to see if it's needed

data/calculated_parser/functions.py
NoahGallant-MUN: Fair enough. Also, based on your overall comment. This is a smaller function I made later to extract some of the work instead of rewriting it.
I can take this function and update the initial calculations I wrote without it

data/calculated_parser/functions.py
NoahGallant-MUN: No

data/calculated_parser/lexer.py
NoahGallant-MUN: Ohhh... I guess i've been spelling it wrong all my life

data/nemo.py
NoahGallant-MUN: Good catch, definitely would have broken something

plotting/sound.py
NoahGallant-MUN: I think I took a different approach in later calculations.
I created a function which finds the index, and that allows you to find both the soniclayerdepth, and value in a single line

plotting/sound.py
NoahGallant-MUN: I'm not entirely sure what you mean or what your referring to?

NoahGallant-MUN:
I commented on your comments, if you want to take a look and see what you still want changed that would be good. Some of them don't require another look and I'll change them on Monday.

Overall I don't think you had any serious concerns?

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+            idx = np.where(speed[x,y] == min_val)
+            if (np.isnan(min_val)):
+                speed[x,y] = np.nan
+            elif idx[0].shape[0] > 1:
+                idx = idx[0][0]
+
+                speed[x,y] = depth.values[idx]
+            else:
+                speed[x,y] = depth.values[idx]
+            #speed[x][y] = depth[np.where(speed[x][y] == np.nanmin(speed[x][y]))]  #np.nanmin(speed[x][y])
```
nsoontie: I wonder if lines 88 to 108 can be replaced with
```python
minspeed = np.nanmin(speed, axis=0)

return minspeed
```
https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanmin.html
* If we use this, we are assuming the speed array always has depth as the first dimension. 
* If speed has dimensions [z,y,x]  then minspeed should have dimensions [y, x]
* This would give a runtime warning if a position has nans over the whole depth and then return a nan for that position

data/calculated_parser/functions.py
nsoontie: If we know the speed is shape [z,y,x] then we can avoid the transpose by iterating like this:
```python
sld = np.nan((speed.shape[-2], speed.shape[-1]))
for x in range(speed.shape[-1]):
   for y in range(speed.shape[-2]):
      sca_value = np.nanmin(speed[:, y, x])
      ...
      sld[y, x] = ...
return sld
```

But, it would take more thought to make sure it works for speed arrays that are shape [t,z,y,x]  if this scenario is possible. ( I'm not sure how the time series functionality works in the calculation layer). In fact, I'm not sure the transpose works for arrays shape [t,z,...] either. Has that been tested?

From what I see the transpose moves the depth dimension from the beginning to the end, for example [z,y,x] to [y,x,z].

data/calculated_parser/functions.py
```diff
@@ -69,11 +70,293 @@ def sspeed(depth, latitude, temperature, salinity):
...
+                    sld_idx = find_sld_idx(sca_idx, speed_point)
+
+                    if not np.isnan(sld_idx):
+                        cd_idx = find_cd_idx(sca_idx, sld_idx, speed_point)
+                        if not np.isnan(cd_idx):
+                            #cd_value = speed_point[cd_idx]
+                            #cd_depth = depth.values[cd_idx]
+
+                            # Now that we have the nearest critical depth idx we must perform linear interpolation
+                            cd_depth = cd_interpolation(cd_idx, sld_idx, speed_point, depth)
```
nsoontie: Here we interpolate between the sonic layer depth and the depth of the grid point closest to the critical depth
Shouldn't we do something like this:

``` python
# If a critical depth exists, speed should be increasing with depth ( right??)
# So if speed_point[cd_indx] < sld_value -> We need to look deeper to cross critical depth - use cd_idx+1
# if speed_point[cd_idx] > sld_value - We need to look shallower to cross critical_depth - use cd_idx - 1
if  speed_point[cd_idx] < speed_point[sld_idx]
    cd_depth = cd_interpolation(cd_idx, cd_idx+1, speed_point, depth) # Question - can we hit a nan here??
else
    cd_depth = cd_interpolation(cd_idx,  cd_idx-1, speed_point, depth)
```

I think it would be more accurate to interpolate around the neighbourhood of the depth closest to the critical layer depth, rather than look all the way back up to the sonic layer depth. But I need to check my assumption that the sound speed is increasing at this point...

We can chat more next week.

NoahGallant-MUN: I think this is what I'm doing, I'm not using the critical depth, just the sound speed at that value.
I'm looking for the depth at which the sonic layer depths sspeed occurs by looking at the depth levels before and after the expected point, and performing a linear interpolation with those 2 depth levels.

Let me know if I misunderstood.

nsoontie: Ok, nevermind... I took a closer look at the interpolation function and you do what I suggested. My bad!!

data/calculated_parser/functions.py
nsoontie: This functions does something different than what the doscstrings  describe. It finds the depth of the minimum speed of sound (not the minimum speed of sound)

So ignore my comment above! I would suggest a more appropriate function name and description.

`np.nanargmin` might be useful here
https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanargmin.html

data/calculated_parser/functions.py
```diff
@@ -9,6 +9,7 @@
 import xarray as xr
 from metpy.units import units
 from pint import UnitRegistry
+import time
```
htmlboss: Is anything from `time` being used in this file?

plotting/sound.py
htmlboss: Instead of `# ~~`, would a new line be sufficient to increase legibility?

data/calculated_parser/parser.py
```diff
@@ -69,8 +66,36 @@ def get_key_for_variable(self, variable):
                 key = [d[k] for k in variable.dimensions]
         except KeyError:
             raise SyntaxError
+        return tuple(key)
+
+    def get_key_for_var_all(self, variable):
```
htmlboss: A docstring would be suitable here I think.

data/calculated_parser/lexer.py
htmlboss: Better late than never 🤣 https://en.wikipedia.org/wiki/Tilde

htmlboss:
> I commented on your comments, if you want to take a look and see what you still want changed that would be good. Some of them don't require another look and I'll change them on Monday.
> 
> Overall I don't think you had any serious concerns?

Nope. Overall it's good. Just those nit-picky things I identified. I would like to see unit tests for the new functions that you wrote AFTER we're all happy with the validity of the results.

data/calculated_parser/functions.py
NoahGallant-MUN: removed

data/calculated_parser/functions.py
NoahGallant-MUN: So I do remember the transpose, and it was important for the way I was doing it. However, Nancy's solution also works. I will convert everything over to using this method.

plotting/sound.py
NoahGallant-MUN: Personally I think it's more clear with a visible barrier.
If you want it gone though I can remove it? I don't really care either way.

plotting/sound.py
NoahGallant-MUN: Added a note to the docstring to explain this

data/calculated_parser/functions.py
NoahGallant-MUN: Removed

data/calculated_parser/functions.py
NoahGallant-MUN: removed double transpose using nancy's solution

data/calculated_parser/functions.py
NoahGallant-MUN: Fixed

data/calculated_parser/functions.py
NoahGallant-MUN: Removed double transpose

plotting/sound.py
NoahGallant-MUN: So I changed all 4 derived products to use the find_###_idx() functions I created.
Does this satisfy your concerns? Often times the value is not needed.

data/calculated_parser/functions.py
NoahGallant-MUN: Changed function name to soundchannelaxis()

NoahGallant-MUN:
> > I commented on your comments, if you want to take a look and see what you still want changed that would be good. Some of them don't require another look and I'll change them on Monday.
> > Overall I don't think you had any serious concerns?
> 
> Nope. Overall it's good. Just those nit-picky things I identified. I would like to see unit tests for the new functions that you wrote AFTER we're all happy with the validity of the results.

Yes, I can definitely add unit tests if you tell me how to run them :)   (I'll write it down somewhere this time)

data/calculated_parser/parser.py
NoahGallant-MUN: Updated docstring, can you take a look and let me know if it's sufficient.
I just referenced the above function and made note of the difference (doesn't perform depth slicing)

data/calculated_parser/functions.py
NoahGallant-MUN: Removed double transpose

data/calculated_parser/functions.py
NoahGallant-MUN: Added to docstring for sca, sld, and cd

data/calculated_parser/functions.py
NoahGallant-MUN: removed

data/calculated_parser/functions.py
NoahGallant-MUN: fixed

data/calculated_parser/lexer.py
NoahGallant-MUN: fixed

data/calculated_parser/functions.py
```diff
@@ -69,11 +69,313 @@ def sspeed(depth, latitude, temperature, salinity):
...
+
+                    if (np.isnan(sld_idx)):
+                        speed[:,y,x] = sld_idx
+                    else:
+                        sld = depth.values[sld_idx]
+                        speed[:, y, x] = sld
+            else:
+                speed[:,y,x] = np.nan
+
+    return speed[0] # Only return one horizontal slice
```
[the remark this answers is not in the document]
htmlboss: From the meeting I would only suggest altering the first bullet I discussed.

data/calculated_parser/functions.py
```diff
@@ -69,11 +69,313 @@ def sspeed(depth, latitude, temperature, salinity):
...
+    Finds the idx (depth layer) of the Sound Channel Axis
+
+    Parameters:
+    speed: np.array of sound speed values in a single point profile
+
+    note: sca - Sound Channel Axis
+    """
+    
+    sca_value = np.nanmin(speed)
+    idx = np.where(speed == sca_value)
```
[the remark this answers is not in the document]
NoahGallant-MUN: Didn't know that was a thing. Cool!

data/calculated_parser/functions.py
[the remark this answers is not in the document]
nsoontie: I agree with Nabil on the first bullet. I was confused by modifying the speed array on my first pass through the code. It would be more readable to create a new array and add to that. 

On the second bullet, I think it will take some effort to vectorize and that should come in a second revision. When we get there, we might consider functions like numpy.nanargmin. My concern is that this will through a ValueError if it encounters a slice with all NaNs which could happen over land.
https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanargmin.html

data/calculated_parser/functions.py
```diff
@@ -69,11 +69,313 @@ def sspeed(depth, latitude, temperature, salinity):
...
+
+    if np.isnan(sca_value):
```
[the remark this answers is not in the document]
NoahGallant-MUN: Already moved it

data/calculated_parser/functions.py
```diff
@@ -69,11 +69,313 @@ def sspeed(depth, latitude, temperature, salinity):
...
+    depth: The depth(s) in meters
+    lat: The latitude(s) in degrees North
+    temperature: The temperatures(s) (at all depths) in celsius
+    salinity: The salinity (at all depths) (unitless)
+    """
+    
+    speed = sspeed(depth, lat, temperature, salinity)
+    for x in range(speed.shape[-1]):
+        for y in range(speed.shape[-2]):
+            if (speed[:,y,x].size - np.count_nonzero(np.isnan(speed[:,y,x]))) != 0:
```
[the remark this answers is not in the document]
NoahGallant-MUN: Done

data/calculated_parser/functions.py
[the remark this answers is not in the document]
NoahGallant-MUN: Partially done (the part that was requested)

data/calculated_parser/functions.py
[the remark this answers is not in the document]
NoahGallant-MUN: Not done because of nancy's concerns

data/calculated_parser/functions.py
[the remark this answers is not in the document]
NoahGallant-MUN: Created a new issue for vectorization

PR #514, opened 2019-10-24
Revert "Sound Channel Calculations"

Ocean-Navigator:
Reverts DFO-Ocean-Navigator/Ocean-Data-Map-Project#510


htmlboss:
Why was this reverted?

PR #533, opened 2019-10-31
Add Unittests for Sound Channel Calculations and SSCP

NoahGallant-MUN:
Adds Unittests for the point functions for each of the sound speed calculations.
This means that it's testing that the correct value is being selected from the profile, but is not testing the looping through each profile.

This also adds sscp which is the possibility of a secondary sound channel at that point, and is either 0 if their isn't a possibility, or 1 if their is a possibility.
(This also has tests)

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -334,8 +334,9 @@ def slopeofsomething(depth, lat, temperature, salinity):
     for x in range(speed.shape[-1]):
         for y in range(speed.shape[-2]):
             speed_point = speed[:,y,x]
-            if count_numerical_vals(speed_point) != 0:
-                slope = slopeofsomething_point(speed_point, depth)
+            num = count_numerical_vals(speed_point)
+            if num != 0:
+                slope = slopeofsomething_point(speed_point[:num], depth[:num])
                 result[y,x] = slope
 
     return result
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -244,7 +244,9 @@ def sscp_point(sspeed, max_idx):
     Ensures:
     Returns either np.nan or 1
     """
-    channels = []
+    
+    print(something)
+
     for d in range(sspeed.shape[0]):
         if d is not 0 and d is not sspeed.shape[0] -1:
             idx = __is_min(sspeed[d-1], sspeed[d], sspeed[d+1])
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -248,9 +248,10 @@ def sscp_point(sspeed, max_idx):
     
     mins = argrelextrema(sspeed, np.less)
     if len(mins) > 1:
+        print(something)
         return 1
     else:
-        return np.nan
+        return 0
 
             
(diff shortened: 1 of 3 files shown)


htmlboss:
When creating PRs please ensure the assignes, labels, and project fields are filled out. It helps with tracking things down the road.

tests/test_functions.py
```diff
@@ -0,0 +1,162 @@
...
+        # Testing 2 sound channels, 2 critical depths
+        sspeed_4 = np.array([1435.76, 1434.12, 1434.98, 1436.27, 1434.51, 1433.97, 1433.49, 1433.67, 1435.09, 1435.98, 1436.12, 1437.41, 1438.19])
+        sca_idx = 6
+        sld_idx = 3
+
+        test_result = find_cd_idx(sca_idx, sld_idx, sspeed_4)
+        self.assertEqual(test_result, 10)
+        
+
+    #def test_cd_interpolation(self):
```
htmlboss: I think this decorator would help here: `@unittest.skip("reason for skipping")`. https://stackoverflow.com/a/16138561/2231969 so then you won't need to un-comment every line individually.

tests/test_functions.py
```diff
@@ -0,0 +1,162 @@
...
+
+        #sspeed_3 = np.array([1437.32, 1436.70, 1435.45, 1433.97, 1436.16, 1435.27, 1433.81, 1433.11, 1433.79, 1434.98, 1436.29, 1437.16, 1438.12])
+        #sld_idx = 3
+        #cd_idx = 10
+
+        
+        #sspeed_4 = np.array([1435.76, 1434.12, 1434.98, 1436.27, 1434.51, 1433.97, 1433.49, 1433.67, 1435.09, 1435.98, 1436.12, 1437.41, 1438.19])
+        
+
+    #def test_soniclayerdepth(self):
```
htmlboss: `@unittest.skip("reason for skipping")`

tests/test_functions.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+    #def test_criticaldepth(self):
```
htmlboss: `@unittest.skip("reason for skipping")`

tests/test_functions.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+    #def test_depthexcess(self):
```
htmlboss: @unittest.skip("reason for skipping")`

data/calculated_parser/functions.py
```diff
@@ -226,6 +227,131 @@ def linearInterp(x1, y1, x2, y2, x):
 
     return cd_depth
 
+
+def __is_min(p1, p2, p3):
```
htmlboss: Is this function being used?

tests/test_functions.py
NoahGallant-MUN: Good to know, thanks.
Will also keep that in mind for future pr's

data/calculated_parser/functions.py
NoahGallant-MUN: No, switched to using scipy extrema function

PR #536, opened 2019-11-04
Document classes in data package

htmlboss:
Our code documentation is fairly poor at the moment, so I spent some time filling in the blanks for some classes in the `data` package.

### DIFF
--- a/data/__init__.py
+++ b/data/__init__.py
@@ -20,10 +20,25 @@ def open_dataset(dataset, **kwargs):
     Opens a dataset.
 
     Determines the type of model the dataset is from and opens the appropriate
-    data object.
+    data object (Nemo, Mercator, Fvcom).
 
     Params:
-    dataset -- Either a string URL for the dataset, or a DatasetConfig object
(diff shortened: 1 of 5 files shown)


data/__init__.py
```diff
@@ -59,7 +75,7 @@ def open_dataset(dataset, **kwargs):
         else:
             return Nemo(url, **args)
 
-    raise TypeError("Dataset url is None.")
+    raise ValueError("Dataset url is None.")
```
htmlboss: This should have been a `ValueError` all along.

data/__init__.py
```diff
@@ -43,7 +58,8 @@ def open_dataset(dataset, **kwargs):
         args = {}
         args['calculated'] = calculated
         args['meta_only'] = __meta_only(**kwargs)
-        args['dataset_key'] = dataset.key
+        if __is_datasetconfig_object(dataset):
```
htmlboss: Adding this if-check prevents an edge-case crash.

PR #542, opened 2019-11-05
Colormap_diff Fix

### LINKED ISSUE #541: Diff. Custom Colourmap Broken
**Describe the bug**
When you change the diff colourmap in a comparison, it doesn't do anything

**To Reproduce**
Do a diff. then change the colourmap

**Expected behavior**
The area plot should reload with the new selected colourmap

**Screenshots**

image

**Additional context**
Erica Discovered this problem

NoahGallant-MUN:
Resolves #541 

Recently, the colormap when drawing an area map for dataset comparison was broken.

This fixes it.

Fixed Plot:
image

### DIFF
--- a/oceannavigator/frontend/src/components/AreaWindow.jsx
+++ b/oceannavigator/frontend/src/components/AreaWindow.jsx
@@ -99,9 +99,7 @@ export default class AreaWindow extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (prevState.colormap_diff !== this.state.colormap_diff) {
-      this.forceUpdate()
-    }
+    
     if (this._mounted && stringify(prevProps) !== stringify(this.props)) {
       if (prevProps.scale !== this.props.scale) {
         if (this.state.scale.indexOf("auto") !== -1) {
@@ -127,12 +125,6 @@ export default class AreaWindow extends React.Component {
     }
   }
 
-  /*componentWillReceiveProps(props) {
-    if (this._mounted && stringify(this.props) !== stringify(props)) {
-
-      
-    }
-  }*/
 
   //Updates Plot with User Specified Title
   updatePlotTitle(title) {
--- a/oceannavigator/frontend/src/components/AreaWindow.jsx
+++ b/oceannavigator/frontend/src/components/AreaWindow.jsx
@@ -103,24 +103,24 @@ export default class AreaWindow extends React.Component {
       this.forceUpdate()
     }
     if (this._mounted && stringify(prevProps) !== stringify(this.props)) {
-      if (props.scale !== this.props.scale) {
+      if (prevProps.scale !== this.props.scale) {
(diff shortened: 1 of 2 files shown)


oceannavigator/frontend/src/components/AreaWindow.jsx
```diff
@@ -98,33 +98,42 @@ export default class AreaWindow extends React.Component {
...
+            output_starttime: this.props.dataset_0.time,
+            output_endtime: this.props.dataset_0.time
           }
         );
       }
     }
   }
 
+  /*componentWillReceiveProps(props) {
+    if (this._mounted && stringify(this.props) !== stringify(props)) {
```
htmlboss: If we're not using this code anymore, might as well delete it. If we need it in the future we can look back in the git history

plotting/map.py
```diff
@@ -344,6 +344,7 @@ def load_data(self):
 
         if self.compare:
             self.variable_name += " Difference"
+            self.cmap = cmap = colormap.find_colormap(self.compare.get('colormap_diff'))
```
htmlboss: I looked further down the file and I can see why you've got the chained assignment thing going on here. What you've is fine for now.

htmlboss:
Another cool thing I learned about github PRs a while ago is that if you put "Resolves ISSUE_NUMER" in the initial PR message, github will automatically close the linked issue when the PR is merged. I've made an edit to your message as an example. Also, when you navigate to that issue, you'll see this PR linked to it at the bottom of the timeline.

plotting/map.py
NoahGallant-MUN: I did not even know you could do that, or that I had done it.
Should I i get rid of the cmap assignment? lol

oceannavigator/frontend/src/components/AreaWindow.jsx
NoahGallant-MUN: Will do

NoahGallant-MUN:
Do we want to do that to attach the pull request? I've been using the 'connect this pull request with an existing issue' option that zenhub gives and it moves it around it's panels.

oceannavigator/frontend/src/components/AreaWindow.jsx
NoahGallant-MUN: > Judging from the change from `componentWillReceiveProps` to `componentDidUpdate`, React wasn't updating the state properly?
> 
> Anyway other than the comment about the unneeded code, it's good to go. Would it also be possible to attach a screenshot of the fixed plot in the original post? I find it helps with constructing a timeline in the event something breaks down the line.

The intention was to fix the reloading issue. It didn't work.
That being said, componentWillReceiveProps throws deprecation warnings, so we should switch over eventually anyway.

plotting/map.py
htmlboss: LOL well I think it's needed since  `cmap` (not `self.cmap`) is referenced below here.

PR #544, opened 2019-11-06
Fix virtual mooring crash

htmlboss:
* The time indexing was wrong. Not sure how I missed that.

Resulting plot:
image

### DIFF
--- a/plotting/timeseries.py
+++ b/plotting/timeseries.py
@@ -109,7 +109,9 @@ class TimeseriesPlotter(PointPlotter):
                 if factor != 1.0:
                     point_data[idx] = np.multiply(point_data[idx], factor)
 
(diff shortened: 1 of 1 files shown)


plotting/timeseries.py
```diff
@@ -109,7 +109,9 @@ def load_data(self):
                 if factor != 1.0:
                     point_data[idx] = np.multiply(point_data[idx], factor)
 
-            times = dataset.timestamps[self.starttime:self.endtime + 1]
```
htmlboss: This was using the netcdf timestamps as indices...which doesn't make sense, so all I had to do was convert these to time indices.

NoahGallant-MUN:
Perhaps in the future, we should perform reviews of everyones code? Even the best can make mistakes :P

PR #559, opened 2019-11-20
Fix loading of Mercator datasets

htmlboss:
## Intro
This issue came up when datasets using the Mercator grid were being requested in the Navigator. The following error was occuring: 
```sh
File "/home/nabil/Ocean-Data-Map-Project/data/nemo.py", line 78, in __bounding_box
    minx, maxx = fix_limits(x, latvar.shape[1])
IndexError: tuple index out of range
```
The error being thrown from `nemo.py` was the obvious giveaway. Why was a mercator-gridded dataset using a nemo file? Turns out the bounding box function had nothing to do with this. The problem had to be in the `open_dataset` logic, so that's where the fixes are.

## Fix
Instead of looking at the dataset variables like we used to be, we now look at the dimensions since they're more consistent across datasets.

The primary observation was that Mercator and Nemo dimension names are always different and more-or-less follow a naming scheme:
Nemo:
image
Mercator:
image

Nemo can vary a bit ('x' or 'xc') but it still has a distinguishable pattern.

## Tests
I've written, fixed, and ran the relevant tests and they pass.

## Result
Here's a screenshot of the fix in action with the Mercator global reanalysis dataset selected:
image

The tiling is messed up for some reason but that's outside the scope of these changes:
Screenshot at 2019-11-20 10-04-26

NOTE: there are also issues with the Timepicker.jsx component that causes the front-end to crash. I've checked the api call results for a few other plots and they're returning 200. Again, fixing the front-end is outside the scope of these changes.

### DIFF
--- a/data/__init__.py
+++ b/data/__init__.py
@@ -51,8 +51,8 @@ def open_dataset(dataset, **kwargs):
 
     if url is not None:
 
-        variable_list = __get_variables(url)
-        if not variable_list:
+        dimension_list = __get_dimensions(url)
+        if not dimension_list:
             raise RuntimeError("Dataset not supported: " + url)
 
         args = {}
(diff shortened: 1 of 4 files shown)


jmunroe:
I agree this fixes the immediate issue so happy to approve.

But, the more general question is why do we have to do this heuristics of trying to guess what kind of model each dataset really is?  Should this not be some property associate with the dataset record itself?

PR #569, opened 2019-12-03
Calculation Vectorization

NoahGallant-MUN:
Vectorized (ie. Improved Efficiency) the Sonci Layer Depth and Sound Channel Axis calculations.

- Benchmarking showed a fairly significant performance improvement.
- Other variables are coming but are proving to be a challenge


data/calculated_parser/functions.py
```diff
@@ -450,7 +459,47 @@ def criticaldepth(depth, lat, temperature, salinity):
     salinity: The salinity (at all depths) (unitless)
     """
 
+    
     speed = sspeed(depth, lat, temperature, salinity)
+
+
+    """
+    #VECTORIZING -- INCOMPLETE
```
htmlboss: Large commented-out code blocks should not be put on master @dwayne-hart

data/calculated_parser/functions.py
```diff
@@ -11,6 +11,7 @@
 from pint import UnitRegistry
 from scipy.signal import argrelextrema
 from scipy.stats import linregress
+import time
```
htmlboss: Is this import being used?

data/calculated_parser/functions.py
```diff
@@ -385,20 +386,29 @@ def soundchannelaxis(depth, lat, temperature, salinity):
...
-            else:
-                result[y,x] = np.nan
-                
-    return result
+    
+    # Mask Array to allow for min index search
+    array = np.ma.masked_array(speed, np.isnan(speed))
+
+    # Find index's of minimum values
+    min_idx = array.argmin(axis = 0)
```
htmlboss: Could `np.nanargmin` be used here?

https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanargmin.html

data/calculated_parser/functions.py
```diff
@@ -417,27 +427,26 @@ def soniclayerdepth(depth, lat, temperature, salinity):
...
-                        result[y,x] = sld_idx
-                    else:
-                        sld = depth.values[sld_idx]
-                        result[y,x] = sld
-            else:
-                result[y,x] = np.nan
 
-    return result # Only return one horizontal slice
+    marray = np.ma.masked_array(speed, np.isnan(speed))
+    min_idx = marray.argmin(axis=0)
```
htmlboss: [`np.nanargmin`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanargmin.html)

data/calculated_parser/functions.py
htmlboss: If this part won't be completed soon, it should be pulled out of this PR and dumped into another branch w/ draft PR.

data/calculated_parser/functions.py
NoahGallant-MUN: No, this returns a value error when it hits an all NaN slice

data/calculated_parser/functions.py
NoahGallant-MUN: No, all NaN slices (as above)

data/calculated_parser/functions.py
NoahGallant-MUN: Moved code to an issue instead

data/calculated_parser/functions.py
NoahGallant-MUN: No I removed it

PR #594, opened 2019-12-17
Datasetconfig

NoahGallant-MUN:
This fixes the Deep Sound Channel Bottom variable

image

image


htmlboss:
Just needs the update from master

PR #596, opened 2019-12-18
Colourmaps

### LINKED ISSUE #585: Default colourmaps for variables are not being displayed
The old colourmaps that changed with the selected variable are now uniform across several variables. Much research has gone into "optimal" colours for displaying particular data which shouldn't be glossed over without discussion. Another discussion is do we even want the colourmaps changed?

Temperature:
Screenshot at 2019-12-17 09-38-43

Water Velocity:
Screenshot at 2019-12-17 09-40-48

Salinity:

NoahGallant-MUN:
Fixed various default colormaps  

resolves #585 

image


htmlboss:
New issue in #597

NoahGallant-MUN:
I agree, I'm not sure a regex pattern would work though since the variable types using the same colourmap could be vastly different.
A datasetconfig entry could work, but I'm hesitant to continuously add more and more to that.

PR #605, opened 2019-12-19
Makes the Cookie disclaimer smaller

NoahGallant-MUN:
The text was too large and causing the scrollbar to appear

image


htmlboss:
Just needs an update from master and it's good to go.

PR #613, opened 2020-01-09
Changed default dataset ID

NoahGallant-MUN:
Should Fix the default dataset problem.
## We should really have a better way of identifying a default dataset though ##


htmlboss:
A regex pattern would solve this I think.

PR #614, opened 2020-01-09
Implement LRU cache for open_dataset

htmlboss:
## Intro
Simply re-implementing the LRU cache that was present before the sqlite backend refactor in the `open_dataset` function. The idea is to reduce the overhead of opening the same dataset in quick succession by caching the last 16 used dataset objects.

## Method
This implementation is more complicated than the original one because of the presence of un-hashable types (`list`) in the `**kwargs` of `open_dataset` (read the docstring). The json serialisation recursively handles that. There obviously is a performance cost, but it should be less than going through all the file opening logic (i.e. disk reads).

I've also added a helper decorator `freeze_args` that will non-recursively convert a function's arguments to a `frozendict`, which allows the use of python's built-in `@lru_cache` decorator.

The function decorator is a clean implementation that allows for other areas to make use of an LRU cache as well.

Docstrings have been written for both new decorators.

## Tests
I've written more tests with one specifically checking that the cache is being used, with everything passing.

### DIFF
--- a/data/__init__.py
+++ b/data/__init__.py
@@ -9,12 +9,11 @@ from data.fvcom import Fvcom
 from data.mercator import Mercator
 from data.nemo import Nemo
 from data.sqlite_database import SQLiteDatabase
-from data.utils import find_le, roll_time, get_data_vars_from_equation
-
-# We cannot cache by URL anymore since with the sqlite approach it points to a database
-# and the original cache system wasn't aware which individual NC files were opened.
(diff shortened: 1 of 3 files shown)


htmlboss:
I'm gonna ignore the codefactor complaints since `isinstance` is more robust than `type`: https://stackoverflow.com/a/1549854/2231969

NoahGallant-MUN:
Okay, fair enough

PR #622, opened 2020-01-14
Added timezone to object

NoahGallant-MUN:
This might be a fix to the subsetting issue. Like last time it needs to be tested before actually being merged since I don't have a test environment.


NoahGallant-MUN:
> Would this have to be applied to the other plotting windows as well?

I just took a look. The other TimePicker in the AreaWindow looks fine as well as the line window, the point window however looks like it requires the change on line 602.

Another issue I noticed, which is why the hour picker is showing up, and why I didn't notice the error before, is that giops_fc_2dll is set as an hourly quantum in the dataset config, when it looks like it should actually be daily?

PR #628, opened 2020-01-28
Add test class for calculated functions

htmlboss:
## Background

The calculated functions in the `data/calculated_parser` folder at present have no tests. This gets the ball rolling with tests for `min`, `max`, and `magnitude`.

Simplest approach. I also use the numpy test helpers: https://docs.scipy.org/doc/numpy/reference/routines.testing.html

I've defined a variable called `relative_tolerance` which controls how many decimal places are checked for floating-point equality. It's set to `1e-3` or 3 decimal places which should be sufficient for our needs. 

I feel this should be a "global" constant somewhere for behavioral consistency.

## Screenshot(s)

Not needed here.

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/tests/test_data_functions.py
+++ b/tests/test_data_functions.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+import unittest
+
+import numpy as np
+
+import data.calculated_parser.functions as funcs
+
+
+class TestDataFunctions(unittest.TestCase):
+
+    def setUp(self):
+        self.relative_tolerance: float = 1e-3
+
+    def test_max(self):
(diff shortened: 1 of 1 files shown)


htmlboss:
> Looks good to me.
> Not a huge deal but I'd confirm with Nancy or Fraser about the tolerance at some point see what they think is reasonable. As an Engineer-to-be I'm not sure my input is opinion is very valid (I probably would have rounded to the nearest whole number).
> Also, big fan of the one liner to run all the tests :)

LOL every opinion matters.
Yes you're right the tolerance is a scientific question that @nsoontie @FraserO or @jmunroe could chime in on.

PR #629, opened 2020-01-29
Raise SyntaxError when p_error is called

htmlboss:
## Background
parser.py:
When a syntax error occurs in an equation string, part of the `t` parameter is attempted to be assigned to `None`. Unfortunately that results in another unrelated Python exception that confuses the caller. Instead we explicitly raise a `SyntaxError` exception, along with some useful debug info in the message (see below).

I threw in a new test case to check for the new raising behavior, along with some pylint formatting.

> In preparation for bringing back the derived variables.

Easiest way.
PLY is built around error feedback so I opted to pass along a more sane exception message
to the console:

```
SyntaxError: Syntax error in equation: atan2(0,)...LexToken(RPAREN,')',1,8)
```

The new test is only hitting the built-in math functions since they don't need to be mocked. To enhance the coverage, we could mock some data and try some other stuff (e.g. try `votemper -` instead of the proper `votemper - 273.15`).

## Screenshot(s)

n/a

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/calculated_parser/parser.py
+++ b/data/calculated_parser/parser.py
@@ -1,8 +1,10 @@
-import ply.yacc as yacc
-import numpy as np
+#!/usr/bin/env python
 
-import data.calculated_parser.lexer
+import numpy as np
+import ply.yacc as yacc
+
 import data.calculated_parser.functions as functions
+import data.calculated_parser.lexer
 
 
 class Parser:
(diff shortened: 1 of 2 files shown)


data/calculated_parser/parser.py
```diff
@@ -38,9 +41,10 @@ def parse(self, expression, data, key, dims):
         Returns a numpy array of data.
         """
         self.data = data
-        self.result = np.nan
+        self.result = np.nan  # populated by p_statement_expr()
         self.key = key
         self.dims = dims
+        self.expression = expression
```
htmlboss: Bring in the expression so it can be referenced later if a syntax error occurs.

htmlboss:
Due to time constraints I'm gonna override the review request.

PR #630, opened 2020-01-30
Change shape of CalculatedArray to consider user-defined dims

htmlboss:
## Background

When the calculation layer was implemented, it only considered calculated variables to have the same dimensions as the underlying variables. So 3D calculated variables were only constructed from 3D underlying variables, and the same goes with 2D stuff. The `CalculatedArray` simply looked into the dimensions of the underlying variables and copied the maximum `.shape` values. This is just fine given the above consideration.

The issues arises when we consider 2D calculated variables with 3D underlying variables (e.g. Deep Sound Channel). We cannot simply copy the shapes of the 3D variables since we'll be pulling in depth, when there shouldn't be one. To work around this, I simply made the user-defined `dims` attribute in the dataset config file a requirement for all calculated variables. This allows the `CalculatedArray` to copy the shapes of the dims listed in said attribute. 

When a calculated variable is encountered with no `dims` attribute, an exception is thrown with a specific error message to easily track down the issue:
```
KeyError: 'This variable (sspeed) has no dims attribute defined in datasetconfig.json.'
```

* Unit tests were edited to reflect the change.
* Enabled datasets in datasetconfig.json were updated to reflect the new attribute requirement.

This way involved the minimum amount of code changed, while preserving the behavior of the higher-level plotting functions.

In fact this simplifies the shape calculation considerably, while giving more flexibility.

The changes only affect the calculated variable code path.

**NOTE** The dims attribute is now required when declaring calculated variables in datasetconfig.json.

## Screenshot(s)
No changes in resulting plots with calculated variables, as expected:
Screenshot from 2020-01-30 12-32-31

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/tests/test_calculated_data.py
+++ b/tests/test_calculated_data.py
@@ -41,7 +41,7 @@ class TestCalculatedData(unittest.TestCase):
             'votemper_new': {
                 'equation': 'votemper * 2',
                 'long_name': 'Temperature',
-                'dims': ('time_counter', 'deptht', 'x', 'y'),
+                'dims': ('time', 'depth', 'latitude', 'longitude'),
                 'units': 'degree_C',
                 'valid_min': -273.15,
                 'valid_max': 999.0,
@@ -68,6 +68,7 @@ class TestCalculatedData(unittest.TestCase):
             'votemper': {
                 'equation': 'votemper -273.15',
                 'units': 'degree_C',
+                'dims': ('time', 'depth', 'latitude', 'longitude')
             }
         }
         with CalculatedImpl('tests/testdata/mercator_test.nc',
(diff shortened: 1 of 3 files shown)


tests/test_calculated_data.py
```diff
@@ -104,19 +105,19 @@ class TestCalculatedArray(unittest.TestCase):
...
     def test_static(self):
         dataset = xr.Dataset()
-        array = CalculatedArray(dataset, "3 * 5")
+        array = CalculatedArray(dataset, "3 * 5", [])
         self.assertEqual(array[0], 15)
 
     def test_passthrough(self):
         dataset = xr.Dataset({'var': ('x', [1, 2, 3, 4, 5])})
-        array = CalculatedArray(dataset, "var")
+        array = CalculatedArray(dataset, "var", [])
```
htmlboss: These dims can be empty lists since these tests are not considering the shape property.

PR #632, opened 2020-01-31
Add Density, Heat Capacity, Temp Gradient variables

htmlboss:
## Background
Add Density, Heat Capacity, and Adiabatic Temperature Gradient to GIOPS/RIOPS 3D Polar Stereographic datasets.

We simply expose more functions in the `seawater` package to the calculation layer.

Easiest way.

Deleted the generic "gradient" colour map since it didn't necessarily well represent gradient plots. 

## Screenshot(s)
Screenshot from 2020-01-31 14-16-46

Screenshot from 2020-01-31 14-21-38

Screenshot from 2020-01-31 14-22-58

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -55,6 +55,16 @@ def magnitude(a, b):
     return np.sqrt(a ** 2 + b ** 2)
 
 
+def __calc_pressure(depth, latitude):
+    pressure = []
+    try:
+        pressure = [seawater.pres(d, latitude) for d in depth]
+    except TypeError:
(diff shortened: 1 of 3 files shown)


data/calculated_parser/functions.py
```diff
@@ -55,6 +55,16 @@ def magnitude(a, b):
     return np.sqrt(a ** 2 + b ** 2)
 
 
+def __calc_pressure(depth, latitude):
```
htmlboss: refactor this out to reduce duplication

PR #634, opened 2020-02-02
Fix deprecated uses of @abc.abstractproperty

douglatornell:
## Background
See https://docs.python.org/3.6/library/abc.html#abc.abstractproperty

> Deprecated since version 3.3: It is now possible to use property, ...
> with abstractmethod(), making this decorator redundant.

Keeping code up to date is A Good Thing™.
Python 3.2 that required `@abc.abtractproperty` was EOL-ed in March 2016.

Noticed this as I started to work on ioc/factory refactoring.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/data.py
+++ b/data/data.py
@@ -41,19 +41,23 @@ class Data(object, metaclass=abc.ABCMeta):
     def timestamp_to_iso_8601(self, timestamp):
         pass
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def timestamps(self):
         pass
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def depths(self):
         pass
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def variables(self):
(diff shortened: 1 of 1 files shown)


douglatornell:
Thanks! PyCharm (my IDE) gets the credit for spotting the issue. I just did the research to understand what it was on about, and made the PR.

I agree that it is odd the `2to3` didn't handle it. Not sure why there is no deprecation warning. I paid attention because I was recently reading about how lots of deprecations from 3.3 and 3.4 are going to become errors when 3.9 is released, and the issues that will cause because of people either routinely ignoring the warnings, or the warnings being suppressed for various reasons.

By "Good to go" did you mean for me to merge this PR? I can't because I don't have write access. No big deal - just don't want us both to be waiting on the other for the merge to happen :smiley:

htmlboss:
Ah ok! 
Yes good to go means good to merge. Forgot about the access permissions hehe.

PR #635, opened 2020-02-03
Add BIORYS Global Monthly + Daily Datasets

htmlboss:
## Background
Found a biology reanalysis on Copernicus so we downloaded it. Figured it would be a good upgrade from BIOMER.

It's the only way.

Nope

## Screenshot(s)
Monthly:
Screenshot from 2020-02-03 12-24-12

Daily:
Screenshot from 2020-02-03 12-24-41

## Checks
- [x] I ran unit tests.

```sh
```


htmlboss:
Just need the colour scales to be vetted first before merging.

PR #641, opened 2020-02-06
Refactor open_dataset() for  inversion of control & factory architecture

douglatornell:
## Background
Based on design proposed by @htmlboss 
[DataArchitecture.pdf](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/files/4167254/DataArchitecture.pdf)

Also working towards accessing SalishSeaCast dataset from ERDDAP server.

Mostly just resolving the consequences of refactoring `open_dataset()` so that it constructs a `NetCDFData` instance that handles all of the SQLite and ERDDAP/OPeNDAP operations, and returns a model object derived from a new `Model` ABC.

Also did some compulsive Python 3 and idiomatic clean-up, but tried to minimize that in the initial patch because it is big enough as it is, and I don't want to distract from the design refactoring. More code style patches to come...

* At present, this handles `Mercator` and `Nemo` model instances
* ~It also presently only handles lat/lon datasets; `Mercator.__enter__()` blows up for polarstereo datasets~ Polar stereographic datasets are handled by `Nemo` model instance. So, this now handles all datasets in `datasetconfig.json` as of 13-Feb-2020.
* I file like `CalculatedData` could be merged into `NetCDFData`. Thoughts?
* Code review please!! - Please see 5744725 commit message for some details of the refactoring
* Updated and added unit tests:
  * `test_data_open_dataset` updated
  * `test_calculated_data` updated
  * `test_merc` updates and additions partially done
  * `test_nemo` updates and additions partially done
  * new `test_netcdf_data` unit tests module

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests. -- ~Lots of fail!!~ Only failures are tests I'm working on in `test_merc` and `test_nemo` :smiley: 

```sh
```


tests/test_data_open_dataset.py
```diff
@@ -11,116 +10,122 @@
...
 
-        with open_dataset('tests/testdata/nemo_test.nc') as ds:
-            self.assertTrue(isinstance(ds, Nemo))
+    @patch.object(DatasetConfig, "_get_dataset_config")
+    @patch.object(data.calculated, "CalculatedData")
+    def test_open_dataset_meta_only_returns_mercator_object(
+        self, patch_calculated_data, patch_get_dataset_config
+    ):
+        patch_get_dataset_config.return_value = {
+            "giops": {"url": "tests/testdata/mercator_test.nc", "variables": {}}
```
htmlboss: If these mocks get too crazy/repetitive, feel free to create some fixtures in `tests/testdata`.

douglatornell:
It's not clear to me how to write unit tests for ABCs, or if there is value in it, provided that there are tests for the classes that derive from them.

In the case of `data.data.Data`, all it contains now is `@abstractmethod` stubs.

`data.model.Model` has 5 implemented methods (in additions to some `@abstractmethod` stubs):
* `get_area()`
* `get_path_profile()`
* `get_profile_depths()`
* `get_timeseries_point()`
* `get_timeseries_profile()`

that presently all have unit tests in `test_nemo.py`. Is there value in adding versions of those tests to `test_merc.py`?

tests/test_data_open_dataset.py
douglatornell: Thanks., I'll keep that in mind. In this case I need just enough a config to get through to the real business of `open_dataset()`, so I feel like they are too small to deserve separate fixture files.

data/model.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    def __init__(self, nc_data):
+        self.interp: str = nc_data.interp
+        self.radius: int = nc_data.radius
+        self.neighbours: int = nc_data.neighbours
+
+    @property
+    @abc.abstractmethod
+    def depths(self) -> numpy.ndarray:
```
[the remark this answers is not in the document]
douglatornell: Conceptually that makes sense, but because of `routes.api_v1_0.variables_query_v1_0()` calling `open_dataset(..., meta_only=True)` (above), if `depths` moves to `Data`, then we need an implementation in `NetCDFData`. I tried that, moving the implementation from `Mercator` to `NetCDFData`, and it fails with `AttributeError` because `self._dataset.coords` isn't defined yet in that `meta_only=True` case.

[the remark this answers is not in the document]
htmlboss: Right and that goes back to me looking into `meta_only`. I'll tackle that today.

htmlboss:
> It's not clear to me how to write unit tests for ABCs, or if there is value in it, provided that there are tests for the classes that derive from them.
> 
> In the case of `data.data.Data`, all it contains now is `@abstractmethod` stubs.
> 
> `data.model.Model` has 5 implemented methods (in additions to some `@abstractmethod` stubs):
> 
>     * `get_area()`
> 
>     * `get_path_profile()`
> 
>     * `get_profile_depths()`
> 
>     * `get_timeseries_point()`
> 
>     * `get_timeseries_profile()`
> 
> 
> that presently all have unit tests in `test_nemo.py`. Is there value in adding versions of those tests to `test_merc.py`?

From a compiled language standpoint, abstract classes aren't unit-tested since they cannot be instantiated on their own. I think we can safely apply that thinking here since `data.data.Data` contains only abstract methods.

I think adding Mercator-specific tests would be prudent just because there are some differences in how the `__bounding_box` and `__resample` methods are implemented between `Nemo` and `Mercator` which would affect the resulting data returned from the `get_*` methods.

data/model.py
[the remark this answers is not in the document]
htmlboss: Right so I don't really see an easy way to dump `meta_only` at the moment. My thinking is once this refactor is done, it may be easier to tease it out.

data/__init__.py
```diff
@@ -27,6 +17,7 @@ def open_dataset(dataset, **kwargs):
     Params:
         * dataset -- Either a string URL for the dataset, or a DatasetConfig object
 
+    TODO: Not accurate - fix; see routes/api_v1_0.py:118 open_dataset(config, meta_only=True)
```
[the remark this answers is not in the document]
htmlboss: I took a look and I don't think `meta_only` can be easily dropped before the refactor is finished. Let's leave it be for now along with the `depths` oddity and tackle it later.

data/calculated.py
```diff
@@ -24,7 +21,7 @@ def __init__(self, url: str, **kwargs):
         Keyword Parameters:
         calculated -- a dict of the calculated variables.
         """
-        super(CalculatedData, self).__init__(url, **kwargs)
+        super().__init__(url, **kwargs)
```
htmlboss: This `2to3` script really sucks at it's job lol...
The `super()`-with-arguments is a Python 2 relic for anyone interested: https://stackoverflow.com/a/17509938/2231969

tests/test_netcdf_data.py
```diff
@@ -0,0 +1,194 @@
...
+            self.assertEqual(date_formatted, 1)
+
+    def test_convert_to_timestamp_list(self):
+        with NetCDFData("tests/testdata/nemo_test.nc") as nc_data:
+            date_formatted = nc_data.convert_to_timestamp(
+                "2014-05-17T00:00:00Z, 2014-06-16T12:00:00Z"
+            )
+            ## TODO: Should there really be a leading space???
+            expected = {" 2014-06-16T12:00:00Z": 1, "2014-05-17T00:00:00Z": 0}
+            self.assertEqual(date_formatted, expected)
```
douglatornell: @htmlboss This new test shows that `convert_to_timestamp()` includes whitespace around date/time string keys in the dict it returns. My guess is that's not intentional, but you may know best. It is easily prevented with a `.strip()` in the dict item construction.

data/__init__.py
```diff
@@ -1,213 +1,75 @@
...
-    return dimension_list
+        calculated_vars = {}
+
+    if url is None:
+        raise ValueError("Dataset url is None.")
+
+    args = {
+        "calculated": calculated_vars,
+        "meta_only": kwargs.get("meta_only", False),
+        "grid_angle_file_url": getattr(dataset, "grid_angle_file_url", ""),
```
htmlboss: `getattr` I like that.

htmlboss:
Question about the "old style" vs "new style" for class definition in python: I know in 3, we inherit from `object` by default, but what do you think of inheriting explicitly? I bring this up because one of the bots was complaining about declaring property methods in an old-style class: https://lgtm.com/rules/10030086/

They mention portability being an advantage for explicit inheritance but I kinda want a second educated opinion.

douglatornell:
> Question about the "old style" vs "new style" for class definition in python: I know in 3, we inherit from `object` by default, but what do you think of inheriting explicitly? I bring this up because one of the bots was complaining about declaring property methods in an old-style class: https://lgtm.com/rules/10030086/
> 
> They mention portability being an advantage for explicit inheritance but I kinda want a second educated opinion.
> 
> I have no preference either way yet.

IMO, this is a weird one from LGTM. I looked at it pretty hard when it first got flagged. I think it is being triggered by the use of the `metaclass` arg in:
```python
class Data(metaclass=abc.ABCMeta):
```
but the Python 3 docs say that is okay: https://docs.python.org/3/library/abc.html#abc.ABC and that it perhaps help to avoid metaclass conflicts in the case of multiple inheritance. I stuck with `metaclass` in the signature because I am new to the team and the codebase, to minimize the diff,  and because I assumed that someone had made an informed decision about using it. (Aside: Now that I am less new, I would argue pretty strenuously against the use of multiple inheritance generally, for the record :smile: )

I almost want to say that LGTM is bugged in this case. I'm pretty sure that we could make it happy though by changing to:
```python
from abc import ABC

class Data(ABC):
```

As to the general story of old-style vs. new-style classes, that was a Python 2 thing. All classes in Python 3 are new-style, but inheriting from `object` (the signature of a Python 2 new-style class) was retained as a no-op. Since we are happily Python 3, and we like modern code :smile: our class signatures should be:
```python
class Foo:
```
rather than:
```python
class Foo(object):
```
(Can you tell that I am feeling less new, and more willing to express strong opinions :smiling_imp: Feel free to push back when you see fit.)

data/__init__.py
```diff
@@ -1,213 +1,75 @@
...
+    }
+    try:
+        args["dataset_key"] = dataset.key
```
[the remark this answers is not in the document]
douglatornell: Good catch! Except I think it should be `"dataset_key": getattr(dataset, "key", "")` because `DatasetConfig` constructor says `self._dataset_key: str = dataset`.

Done in 6b974df5

data/__init__.py
```diff
@@ -1,213 +1,75 @@
...
+        pass
+
+    nc_data = CalculatedData(url, **args)
+    if not args["meta_only"]:
+        # Get required NC files from database and add to args
+        nc_data.get_nc_file_list(dataset, **kwargs)
+
+    dimension_list = nc_data.dimensions
+    if not dimension_list:
+        raise RuntimeError("Dataset not supported: %s." % url)
```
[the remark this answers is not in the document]
douglatornell: For sure! Dunno how I missed that one :laughing: 

Done in 87bc0377845b

data/netcdf_data.py
```diff
@@ -114,19 +121,26 @@ def timestamp_to_time_index(self, timestamp):
 
         return result if result.shape[0] > 1 else result[0]
 
-    def timestamp_to_iso_8601(self, timestamp):
+    def timestamp_to_iso_8601(self, timestamp: Union[int, List]):
+        """Converts a given timestamp (e.g. 2031436800) or list of timestamps
+        into the corresponding time index(es) for the time dimension.
```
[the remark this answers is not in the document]
douglatornell: Yep. Copy-paste-forgot-to-update :roll_eyes: 

Done in d547b6ed112

tests/test_netcdf_data.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def test_latlon_variables_not_found_raises(self):
+        # Use salishseacast_ssh_test.nc here because it has neither nav_lat nor latitude variables
+        with NetCDFData("tests/testdata/salishseacast_ssh_test.nc") as nc_data:
+            with self.assertRaises(KeyError):
+                lat, lon = nc_data.latlon_variables
+
+    def test_xarray_variables(self):
+        with NetCDFData("tests/testdata/mercator_test.nc") as nc_data:
+            vars = nc_data.variables
```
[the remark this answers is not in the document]
douglatornell: Sorry. Dunno why I didn't notice that...

Done in f3ba4dca

tests/test_netcdf_data.py
```diff
@@ -0,0 +1,230 @@
...
+            self.assertEqual(vars[0].unit, "Kelvin")
+            self.assertEqual(
+                vars[0].dimensions, ("depth", "latitude", "longitude", "time")
+            )
+            self.assertEqual(vars[0].valid_min, 173.0)
+            self.assertEqual(vars[0].valid_max, 373.0)
+
+    def test_fvcom_variables(self):
+        with NetCDFData("tests/testdata/fvcom_test.nc") as nc_data:
+            vars = nc_data.variables
```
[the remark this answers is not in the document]
douglatornell: Done in f3ba4dca

tests/test_netcdf_data.py
```diff
@@ -0,0 +1,230 @@
...
+            self.assertEqual(
+                vars[3].dimensions, ("time", "maxStrlen64", "node", "siglay")
+            )
+            self.assertIsNone(vars[3].valid_min)
+            self.assertIsNone(vars[3].valid_max)
+
+    def test_variable_list_cached(self):
+        with NetCDFData("tests/testdata/nemo_test.nc") as nc_data:
+            self.assertIsNone(nc_data._variable_list)
+            vars = nc_data.variables
```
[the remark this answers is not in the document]
douglatornell: Done in f3ba4dca

douglatornell:
> Overall this is really good. I'm elated to see how the proposed refactor worked so well with the code. tada balloon
> Anyway back to business:
> Partial review...only one issue:
> 
>     * Subsets throwing exceptions (http error 500): can't see from the `NetCDFData.subset` function where that's happening.
>       Example request: http://navigator.oceansdata.ca:81/api/v1.0/subset/?&output_format=NETCDF4&dataset_name=giops_day&variables=votemper&min_range=44.27667127377518,-54.527343213558204&max_range=52.321910885947716,-42.398436963558204&time=2214907200,2214907200&user_grid=0&should_zip=0
> 
> 
> Replacing port 81 with 80 (production) results in a file appearing for download.
> 
> I'll continue reviewing tomorrow.

Fixed in 7359e14e3

Unfortunately, this is a common bug pattern for this refactoring. :cry:  A bunch of attrs & methods that used to be accessible on the instance returned by `open_dataset.__enter__()` moved on to `NetCDFData`, so they have to be accessed via the `nc_data` attr.

Some of those bugs should be caught by `test_api_v_1_0` (as this one was), but I have trouble seeing them because I get a lot of false fails from that module because I can't connect to the THREDDS server. Looking harder at other fails in that module now to try to see if they are being caused by this refactoring.

data/mercator.py
```diff
@@ -1,60 +1,64 @@
...
+from data.netcdf_data import NetCDFData
 from utils.errors import APIError
 
 
-class Mercator(CalculatedData):
+class Mercator(Model):
     __depths = None
 
-    def __init__(self, url, **kwargs):
+    def __init__(self, nc_data: Union[CalculatedData, NetCDFData]) -> None:
```
htmlboss: Since `__init__` doesn't explicitly return anything (https://help.semmle.com/wiki/display/PYTHON/__init__+method+returns+a+value), does it make sense to annotate the return type?

data/mercator.py
```diff
@@ -1,60 +1,64 @@
...
+        super().__init__(nc_data)
         self.latvar = None
         self.lonvar = None
-
-        super(Mercator, self).__init__(url, **kwargs)
+        self.nc_data = nc_data
+        self._dataset = nc_data.dataset
```
htmlboss: I think down the road we should dump `self._dataset` and simply change all references to `self.nc_data.dataset`. I think it'd help cut down on confusion. For now though this is ok.

douglatornell: 100% agree! I was exploring that last weekend and found that it ripples out into a lot of files beyond those touched by this PR, so I decided to wait and do it as a separate task.

data/mercator.py
```diff
@@ -132,13 +136,31 @@ def __resample(self, lat_in, lon_in, lat_out, lon_out, var, radius=50000):
...
 
-            output = super(Mercator, self)._interpolate(input_def, output_def, data.transpose())
+            output = self.nc_data.interpolate(input_def, output_def, data.transpose())
 
         if len(origshape) == 4:
             output = output.reshape(origshape[2:])
 
         return np.squeeze(output)
 
+    def get_path(self, path, depth, time, variable, numpoints=100, times=None, return_depth=False):
```
htmlboss: return-type annotation missing: `np.ndarray`.

data/mercator.py
```diff
@@ -287,15 +308,14 @@ def get_point(self, latitude, longitude, depth, timestamp, variable,
 
         if return_depth:
             return res, dep
-        else:
-            return res
+        return res
 
     def get_profile(self, latitude, longitude, timestamp, variable):
```
htmlboss: return-type annotation missing: `np.ndarray`.

data/nemo.py
```diff
@@ -164,19 +177,37 @@ def __latlon_vars(self, variable):
...
                     return (
-                        self.get_dataset_variable(p[0]),
-                        self.get_dataset_variable(p[1])
+                        self.nc_data.get_dataset_variable(p[0]),
+                        self.nc_data.get_dataset_variable(p[1])
                     )
 
         raise LookupError("Cannot find latitude & longitude variables")
 
+    def get_path(self, path, depth, time, variable, numpoints=100, times=None, return_depth=False):
```
htmlboss: return-type annotation missing: `np.ndarray`.

data/nemo.py
```diff
@@ -164,19 +177,37 @@ def __latlon_vars(self, variable):
...
+        if return_depth:
+            result, dep = self.get_point(lat, lon, depth, time, variable,
+                                         return_depth=return_depth)
+            return np.array([lat, lon]), distances, times, result, dep
+        else:
+            result = self.get_point(lat, lon, depth, time, variable,
+                                    return_depth=return_depth)
+            return np.array([lat, lon]), distances, times, result
+
     def get_raw_point(self, latitude, longitude, depth, timestamp, variable):
```
htmlboss: return-type annotation missing: `np.ndarray`.

data/nemo.py
```diff
@@ -320,11 +351,10 @@ def get_point(self, latitude, longitude, depth, timestamp, variable,
 
         if return_depth:
             return res, dep
-        else:
-            return res
+        return res
 
     def get_profile(self, latitude, longitude, timestamp, variable):
```
htmlboss: return-type annotation missing: `np.ndarray`.

data/netcdf_data.py
```diff
@@ -569,7 +578,27 @@ def latlon_variables(self):
         )
 
     @property
-    def depth_dimensions(self):
+    def dimensions(self) -> List[str]:
```
htmlboss: How we're doing the dimensions here feels a little code-smelly...to call this you need an instance of this class which opens the database or hits the url at start, but this implementation is re-opening all the stuff anyways.

We can address this at a later date so this is good for now.

data/mercator.py
douglatornell: I wondered about that, but [PEP-484 says explicitly](https://www.python.org/dev/peps/pep-0484/#id13) that "the return type of `__init__` ought to be annotated with `-> None`" and provides a rationale.

data/model.py
```diff
@@ -0,0 +1,115 @@
...
+        self,
+        area,
+        depth,
+        time,
+        variable,
+        interp,
+        radius,
+        neighbours,
+        return_depth=False,
+    ):
```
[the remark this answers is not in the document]
douglatornell: I looked at these when I created `Model`. Some, like this `get_area()` one are easy, but others are like `Tuple[]` for `get_path_profile()`, or whatever the notation is for 1 value or a tuple in `get_point()`.

I'll admit that I was lazy and didn't dig through all of the details to figure out all of these annotations. I think that you have confirmed my perception that improving type annotations is a goal, not a requirement. So, as you suggest, I'm going to leave these for now.

douglatornell:
I agree that this is ready to merge.

I've been looking at the in-object caching, and performance, and I think there may be an easy win re: `self.time_variable` and `self.timestamps` in the `__enter__()` methods of `Mercator` and `Nemo`, but I want to do some more testing, and make that a separate PR.

I think that when you merge this we should deploy it on the port :81 container for more testing and tuning before pushing it public. But public by mid-next-week seems possible to me.

htmlboss:
Another pesky error with the subset:
```sh
[2020-03-16 09:02:32 -0300] [9707] [ERROR] Error handling request /api/v1.0/subset/?&output_format=NETCDF4&dataset_name=giops_day&variables=votemper&min_range=44.27667127377518,-54.527343213558204&max_range=52.321910885947716,-42.398436963558204&time=2214907200,2214907200&user_grid=0&should_zip=0
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 135, in handle
    self.handle_request(listener, req, client, addr)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 176, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sdfo000/ocean-nav/miniconda/3/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/buildadm/ocean-nav/Ocean-Data-Map-Project-20200316/routes/api_v1_0.py", line 437, in subset_query_v1_0
    with open_dataset(config, variable=variables, timestamp=int(time_range[0]), endtime=int(time_range[1])) as dataset:
  File "/home/buildadm/ocean-nav/Ocean-Data-Map-Project-20200316/utils/decorators.py", line 71, in lru_decorator
    return cached_function(*_args, **_kwargs)
  File "/home/buildadm/ocean-nav/Ocean-Data-Map-Project-20200316/utils/decorators.py", line 61, in func_with_serialized_params
    return func(*_args, **_kwargs)
  File "/home/buildadm/ocean-nav/Ocean-Data-Map-Project-20200316/data/__init__.py", line 66, in open_dataset
    nc_data.get_nc_file_list(dataset, **kwargs)
  File "/home/buildadm/ocean-nav/Ocean-Data-Map-Project-20200316/data/netcdf_data.py", line 748, in get_nc_file_list
    raise RuntimeError("Error finding timestamp(s) in database.")
RuntimeError: Error finding timestamp(s) in database.
```

PR #642, opened 2020-02-10
Extend calculation layer to support full-depth slicing

htmlboss:
## Background
The requested acoustic calculations required the "compression" of 3D data variables into 2D (i.e. full-depth temperature + full-depth salinity would yield a 2D deep sound channel product). The parser grammar didn't support this kind of slicing so we needed to add new tokens ("[" and "]"). 

To request a variable be sliced with full-depth, use the following syntax in the dataset config: `[ my-variable-key ]`. The square brackets tell the parser to grab the full depth dimension when slicing the requested lat/lon coords.

Simplest way I could think of to minimize the amount of code changed, while keeping present functionality. The square brackets were borrowed from math, where they refer to inclusive ranges; where we want the full depth range.

The `dims` attribute for calculated variables in the dataset config remain required courtesy of #630. This partly allows the full-depth slicing implemented here.

I've also specifically tested the subsetting code with no issue.

## Screenshot(s)
I jumped a little ahead and implemented the deep sound channel variable for testing; not in this pr:
image

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -128,26 +128,10 @@ class CalculatedArray():
             pass
 
     def __getitem__(self, key):
-        """This is where the magic happens.
-        The dimensions of the key are determined by looking at the dimensions
-        of all the underlying variables used in the calculation, this is then
-        passed along to the parser where the calculation is performed.
-        """
-        key_dims = ()
-        for v in self._parser.lexer.variables:
-            if v not in self._parent.variables:
-                continue
-
-            d = self.__get_parent_variable_dims(v)
-
-            if len(d) > len(key_dims):
-                key_dims = d
-            elif len(d) == len(key_dims):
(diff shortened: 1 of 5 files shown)


data/calculated.py
```diff
@@ -128,26 +128,10 @@ def __init__(self, parent, expression, dims, attrs={}, db_url=""):
             pass
 
     def __getitem__(self, key):
-        """This is where the magic happens.
-        The dimensions of the key are determined by looking at the dimensions
-        of all the underlying variables used in the calculation, this is then
-        passed along to the parser where the calculation is performed.
-        """
-        key_dims = ()
```
htmlboss: Delete all this stuff since we explicitly define the dims in the dataset config a la #630

PR #643, opened 2020-02-11
Add Deep Sound Channel calculated variable

htmlboss:
## Background
tldr: Someone wanted sound channel calculations, and so we deliver.

Available in:
* GIOPS Forecast 3D (Lat/Lon)
* GIOPS Forecast 3D PS
* RIOPS Forecast 3D PS

Deep Sound Channel is defined as the depth where the minimum sound speed occurs in a given vertical column.

I also added a basic test for the new `deepsoundchannel` function.

Fastest performing implementation since it's vectorized.

Not really. Vectorized code sometimes isn't the most intuitive to understand so I've added some comments to hopefully clarify things.

## Screenshot(s)
Screenshot from 2020-02-11 10-16-39

Screenshot from 2020-02-11 11-17-30

## Bonus
Fixed a crash when calculating pressure with GIOPS Forecast 3D (Lat/Lon) so speed of sound is now available there too.

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import functools
+from typing import Union
 
 import metpy.calc
 import numpy as np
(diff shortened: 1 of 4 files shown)


htmlboss:
Ignore codefactor complaining...numpy doesn't play well with `isinstance`.

PR #648, opened 2020-02-12
Deprecate vectors api argument

htmlboss:
## Background
Extension of #647 where I removed the need for the `vectors` option in the API since we calculate the magnitude properly through the calculation layer.

This resolves the duplicate magnitude variables in the front-end introduced by #647 

Did a search of the code and removed all references of `vectors`.

It's the only way.

Nope

## Screenshot(s)
n/a

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/routes/api_v1_0.py
+++ b/routes/api_v1_0.py
@@ -114,9 +114,15 @@ def variables_query_v1_0():
 
     data = []
 
-    if 'vectors_only' not in args:
+    if 'vectors_only' in args:
+        for variable in config.vector_variables:
+            data.append({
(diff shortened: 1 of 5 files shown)


routes/api_v1_0.py
```diff
@@ -114,9 +114,15 @@ def variables_query_v1_0():
 
     data = []
 
-    if 'vectors_only' not in args:
+    if 'vectors_only' in args:
```
htmlboss: simplified the janky logic

PR #657, opened 2020-02-18
API cleanup and testing

jamesprayner:
## Background
 - This PR finishes the move from API v0.0 to API v1.0.
 - All instances of the old API in `frontend` have been changed.
 - Definitions from `routes_impl.py` have been moved to `api_v1_0.py`.

Tests have been written for every API endpoint. During this process I found a lot of the endpoints don't work as they're supposed to. I have made notes on each of these and will make separate issues for them. As a result of these errors, the tests associated with broken endpoints are flagged to be skipped until they get fixed.

The condition of the API has been unclear for a long time. This new formatting and new tests will make it a lot easier to keep all parts of the API working correctly in the future.

Test scripts and `frontend` changes

## Screenshot(s)
None

## Checks
- [x] I ran unit tests.

```sh
```


routes/api_v1_0.py
```diff
@@ -56,7 +98,28 @@ def datasets_query_v1_0():
...
+    else:
+        for key in DatasetConfig.get_datasets():
+            config = DatasetConfig(key)
+            data.append({
+                'id': key,
+                'value': config.name
+            })
+    data = sorted(data, key=lambda k: k['value'])
+    resp = jsonify(data)
+    resp.headers['Access-Control-Allow-Origin'] = '*'
```
htmlboss: I think this line can be deleted now that we have that catch-all cors handler.

routes/api_v1_0.py
```diff
@@ -56,7 +95,27 @@ def datasets_query_v1_0():
         Response -- Response object containing list of available datasets w/ some metadata.
     """
 
-    return routes.routes_impl.query_datasets_impl(request.args)
+    data = []
+    if 'id' not in request.args:
```
htmlboss: Nit pick: to ditch the `not` keyword could we flip the if-else check? I would argue it's more _pythonic_ 🙄

routes/api_v1_0.py
```diff
@@ -210,35 +274,124 @@ def scale_v1_0(dataset: str, variable: str, scale: str):
...
+
+    return send_file(bytesIOBuff, mimetype="image/png", cache_timeout=MAX_CACHE)
 
 
 @bp_v1_0.route('/api/v1.0/range/<string:dataset>/<string:variable>/<string:interp>/<int:radius>/<int:neighbours>/<string:projection>/<string:extent>/<string:depth>/<int:time>.json')
 def range_query_v1_0(dataset: str, variable: str, interp: str, radius: int, neighbours: int, projection: str, extent: str, depth: str, time: int):
-    return routes.routes_impl.range_query_impl(interp, radius, neighbours, dataset, projection, extent, variable, depth, time)
+    extent = list(map(float, extent.split(",")))
+
+    min, max = plotting.scale.get_scale(
```
htmlboss: CodeFactor has a good point here...could we rename these variables to `min_data_val` and `max_data_val` or something similar? I really don't like interfering with built-in `min` and `max` functions.

routes/api_v1_0.py
```diff
@@ -210,35 +274,124 @@ def scale_v1_0(dataset: str, variable: str, scale: str):
...
+    """
+    API Format: /api/v1.0/class4/<string:q>/<string:class4_id>/
+
+    <string:q>         : forecasts / models (Data Request)
+    <string:class4_id> : ID of the desired class4 - Can be found using /api/class4/
+
+    Returns a list of class4 datapoints for a given day 
+    """
+
+    if class4_id == None:
```
htmlboss: We can shorten this to a more _pythonic_ (lol) expression: `if not class4_id:`

routes/api_v1_0.py
```diff
@@ -251,8 +404,18 @@ def stats_v1_0():
         date = dataset.convert_to_timestamp(query.get('time'))
         date = {'time': date}
         query.update(date)
-
-        return routes.routes_impl.stats_impl(args, query)
+        if query == None:
```
htmlboss: `if not query:`

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
+    """
+
+    cache_dir = current_app.config['CACHE_DIR']
+    f = os.path.join(cache_dir, request.path[1:])
+
+    # Check if the tile/image is cached and send it
+    if _is_cache_valid(dataset, f):
+        return send_file(f, mimetype='image/png', cache_timeout=MAX_CACHE)
+    # Render a new tile/image, then cache and send it
+    else:
```
htmlboss: Unnecessary `else` statement since the `if` contains a `return` statement.  I nit-pick this kinda thing in any language.

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
 
 @bp_v1_0.route('/api/v1.0/tiles/topo/<string:shaded_relief>/<string:projection>/<int:zoom>/<int:x>/<int:y>.png')
 def topo_v1_0(shaded_relief: str, projection: str, zoom: int, x: int, y: int):
-    hull_shade = shaded_relief == 'true'
-    return routes.routes_impl.topo_impl(projection, zoom, x, y, hull_shade)
+    """
+        Generates topographical tiles
+    """
+
+    if shaded_relief == "true":
```
htmlboss: How does `bShaded_relief = bool(shaded_relief)` sound? I know we briefly talked about this but I figure let's re-visit now that we've got some tests.

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
+
+    if zoom > 7:
+        return send_file(shape_file_dir + "/blank.png")
+
+    cache_dir = current_app.config['CACHE_DIR']
+    f = os.path.join(cache_dir, request.path[1:])
+
+    if os.path.isfile(f):
+        return send_file(f, mimetype='image/png', cache_timeout=MAX_CACHE)
+    else:
```
htmlboss: Unnecessary `else` statement since the `if` contains a `return` statement.

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
```
htmlboss: Unnecessary `else` statement since the `if` contains a `return` statement.

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
+    basedir = requestf.rsplit("/", 1)[0]
+
+    # Send blank tile if conditions aren't met
+    if (zoom < 7) or (projection != "EPSG:3857"):
+        return send_file(shape_file_dir + "/blank.mbt")
+
+    # Send file if cached or select data in SQLite file
+    if os.path.isfile(requestf):
+        return send_file(requestf)
+    else:
```
htmlboss: Unnecessary `else` statement since the `if` contains a `return` statement.

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
+        y = (2**zoom-1) - y
+        connection = sqlite3.connect(
+            shape_file_dir + "/{}.mbtiles".format(tiletype))
+        selector = connection.cursor()
+        sqlite = "SELECT tile_data FROM tiles WHERE zoom_level = {} AND tile_column = {} AND tile_row = {}".format(
```
htmlboss: Could we use an f-string here?

routes/api_v1_0.py
```diff
@@ -400,28 +795,150 @@ def timestamps():
...
+            zoom, x, y)
+        selector.execute(sqlite)
+        tile = selector.fetchone()
+        if tile == None:
+            return send_file(shape_file_dir + "/blank.mbt")
+
+        # Write tile to cache and send file
+        if not os.path.isdir(basedir):
```
htmlboss: Would `if os.path.isdir(basedir) is None:` make sense here? Not saying it should be changed...just another way of writing this to consider...

routes/api_v1_0.py
```diff
@@ -430,3 +947,49 @@ def after_request(response):
...
+            modtime = datetime.datetime.fromtimestamp(
+                os.path.getmtime(f)
+            )
+            age_hours = (
+                datetime.datetime.now() - modtime
+            ).total_seconds() / 3600
+            if age_hours > cache_time:
+                os.remove(f)
+                return False
+            else:
```
htmlboss: Unnecessary `else` statement since the `if` contains a `return` statement.

htmlboss:
Tests are good. Only thing left is to create the issues of the broken api stuff.

PR #661, opened 2020-02-20
Reduced required zoom level for vector tiles

jamesprayner:
## Background
Hard to see entire great lakes dataset, moved the vector tiles up one level.

only way

no

## Screenshot(s)
Old:
oldlanks

New:
withdata

## Checks
- [x] I ran unit tests.

```sh
```


htmlboss:
I think we should have a further discussion on the UX side first.

PR #666, opened 2020-02-26
Increase Max Zoom

jamesprayner:
## Background
A new set of Mapbox tiles were created to increase the max zoom level. This PR changes some of the boundaries to allow users to zoom in further and use the new tiles.

only way

## Screenshot(s)
Halifax harbor at second to highest zoom level
z15

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map.jsx
@@ -56,7 +56,7 @@ const MIN_ZOOM = {
 };
 
 const MAX_ZOOM = {
-  "EPSG:3857": 13,
+  "EPSG:3857": 16,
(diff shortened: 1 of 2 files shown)


htmlboss:
Nice I can see Norm's office from that picture.

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -56,7 +56,7 @@ const MIN_ZOOM = {
 };
 
 const MAX_ZOOM = {
-  "EPSG:3857": 13,
+  "EPSG:3857": 16,
```
htmlboss: Zoom index starts at 1.

PR #667, opened 2020-02-26
Front-end: display hours in UTC

htmlboss:
## Background
The hour selector in the front-end was automatically converting UTC to the client browsers time zone. This caused some confusion since the datasets we display are in UTC. Also, the time label says UTC.

Tweak `Timepicker.jsx` to format the hours/minutes in UTC and not the client time zone.

Two line change.

cc @nsoontie 

It's the only way.

The signature for `dateFormat` is:
```ts
dateFormat(
    date?: Date | string | number,
    mask?: string,
    utc?: boolean,
    gmt?: boolean
)
```

## Screenshot(s)
image

image

## Checks
- [x] I ran unit tests (not that we have any front-end tests to run lol).

```sh
```


htmlboss:
For some reason I can't get a screenshot while the dropdown is open...

PR #670, opened 2020-03-01
Added a new script to daemonize runserver.sh. Tweaked oceannavigator.…

dwayne-hart:
…cfg for our systems.

## Checks
- [ ] I ran unit tests.

```sh
```


launch-web-service.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
htmlboss: 👍 proper shebang

oceannavigator/oceannavigator.cfg
```diff
@@ -1,15 +1,15 @@
...
-CLASS4_FTP = "dory-dev.cs.dal.ca"
-OBSERVATION_AGG_URL = "http://localhost:8080/thredds/dodsC/Misc/observations/aggregated.ncml"
+DRIFTER_AGG_URL = "http://10.0.3.56:8080/thredds/dodsC/Drifter/aggregated.ncml"
+DRIFTER_URL = "http://10.0.3.56:8080/thredds/dodsC/Drifter/%s.nc"
+DRIFTER_CATALOG_URL = "http://10.0.3.56:8080/thredds/catalog/Drifter/catalog.xml"
+OPSERVATION_AGG_URL = "http://10.0.3.56:8080/thredds/dodsC/Misc/observations/aggregated.ncml"
+CLASS4_CATALOG_URL = "http://10.0.3.56:8080/thredds/catalog/Class4/catalog.xml"
+CLASS4_URL = "http://10.0.3.56:8080/thredds/dodsC/Class4/%s/%s.nc"
+CLASS4_FTP = "10.0.3.9"
+OBSERVATION_AGG_URL = "http://10.0.3.56:8080/thredds/dodsC/Misc/observations/aggregated.ncml"
```
douglatornell: @dwayne-hart These changes from `localhost` to `10.0.3.56` are pretty unfriendly for my case of running the `test_api_v_1_0` test suite on a remote setup. There are a bunch of test failures either way, but with `localhost` they fail fast. With `10.0.3.56` each test waits ~2min for a timeout on the request to that host before it fails.

I'm not sure what yo suggest as an alternative though because your change is presumably more robust for production.

dwayne-hart: @douglatornell apologies as I was more focused on keeping the production systems' configuration files from being clobbered every time an update occurred. Especially when the current environment is reused instead of going with a directory rename, git clone, yarn compile, and service launch. I have a couple of ideas I may through during next week's call or I may circulate offline through other medium.

dwayne-hart: Many thanks for bringing it forward. We have gotten around this problem by using the following steps;

- git stash
- git fetch && get pull
- git stash apply

this will hide files from being changed when code is updated.

PR #672, opened 2020-03-02
GPSC modifications

dwayne-hart:
## Background
Need to come up with a way such that researchers are able to launch their own Navigator instance. The script will check to see if a network port is free. If it is not it increments by one and checks again. If the port is not  in use it will proceed with launching the Navigator on the available port.

In order to allow multiple users to be able to launch their own instance or possibly multiple instances of the Navigator on various machines.

There is no exponential backoff. The script will just keep going from 5000 by 1 until a free port is found. Also, the user is given the URL which is the <IP Address>:<PORT> where the instance is running.

## Screenshot(s)
Screen Shot 2020-03-02 at 5 36 24 PM

## Checks
N/A

### DIFF
--- a/launch-web-service.sh
+++ b/launch-web-service.sh
@@ -1,3 +1,17 @@
 #!/usr/bin/env bash
+HOST_IP=`ip addr show | egrep "e(th|np)" | grep inet | awk '{print $2}' | sed -e 's/\/[0-9]*//'`
+
+PORT=5000
+
+nc -zv 0.0.0.0 $PORT > /dev/null
+RES=$?
+
+while  [ $((RES)) -eq 0 ] ; do
+	PORT=$((PORT+1))
+	nc -zv 0.0.0.0 $PORT > /dev/null
+	RES=$?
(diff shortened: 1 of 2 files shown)


launch-web-service.sh
```diff
@@ -1,3 +1,17 @@
 #!/usr/bin/env bash
+HOST_IP=`ip addr show | egrep "e(th|np)" | grep inet | awk '{print $2}' | sed -e 's/\/[0-9]*//'`
```
htmlboss: Use $(...) notation instead of legacy backticked `...`. (SC2006)

htmlboss: `egrep` is non-standard and deprecated. Use `grep -E` instead. (SC2196)

dwayne-hart: Implemented the suggested code change.

dwayne-hart: Implemented the suggested code change.

dwayne-hart:
image

PR #678, opened 2020-03-03
Fixed expected value in the 5th decimal place

jmunroe:
There was a 2.27e-5 difference in the expected compared to return value for the profile tests. Updated test to match returned value.

### DIFF
--- a/tests/test_nemo.py
+++ b/tests/test_nemo.py
@@ -49,13 +49,6 @@ class TestNemo(unittest.TestCase):
 
             self.assertEqual(idx, 0)
 
-    def test_timestamp_to_time_index(self):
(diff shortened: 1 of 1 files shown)


tests/test_nemo.py
```diff
@@ -42,12 +42,6 @@ def test_variables(self, mock_query_func):
             self.assertEqual(sorted(variables['votemper'].dimensions), sorted(
                 ["deptht", "time_counter", "y", "x"]))
 
-    def test_timestamp_to_time_index(self):
```
htmlboss: Was deleting this test intended?

jmunroe:
Yes. That test is duplicated (same name, same content)

JustinElms:
Closed because this test is no longer in codebase.

PR #681, opened 2020-03-15
Update to the runserver.sh script

dwayne-hart:
Launch as many gunicorn processes as their are processors. The timeout was increased for both the silent works terminated and restated. And again for the graceful restart of workers.

## Background
Previously we set the number of gunicorn processes to some value. Depending on the instance resources there were either to many or to few gunicorn services running. This will launch the number of gunicorn processes as their are processors. 

Also, the timeout values were increased from 90s to 300s. Depending on what datasets were selected and if there was a large time range requested. The stalled gunicorn processes were terminated.

## Checks
- [ N/A] I ran unit tests.

### DIFF
--- a/runserver.sh
+++ b/runserver.sh
@@ -6,6 +6,6 @@
 
 [[ x"$1" == x"" ]] && PORT=5000 || PORT=$1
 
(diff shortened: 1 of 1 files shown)


dwayne-hart:
Awesome.

PR #683, opened 2020-03-17
Database backed observational data.

geoffholden:
This is the data model only, the frontend, plotting, and API requests are coming later.

## Background
This is the first PR for the observational data work. It's using SQLAlchemy (specifically Flask-SQLAlchemy) for the ORM mapping.

The general idea is that all the observational data in our local DB is organized as follows:

Platforms are individual sampling things (not a type of thing). So an individual drifter is a platform, as is a glider, an argo float, etc. an oceanographic cruise/mission is also considered a platform.
A platform has a type, a unique identifier (so that a data importer can query for that platform again on another day to append more data), and key/value metadata.

Stations are locations where and when data collection happens. They can optionally have a name, and are required to have a latitude, longitude, and time. They are also linked back to an individual platform.

Samples are the basic unit of data. They are linked back to a station, and have a datatype, a depth, and a value.

DataTypes have a key (encouraging the use of CF-Standard names for the keys), a name, and a unit. Every Sample references a DataType.

DB Schema: https://dbdiagram.io/d/5e70e2d14495b02c3b885c95

I took this approach after a few iterations on data models. In this model, all currently communicated classes of observational data fit and can be just treated as 'observational data'. This should remove the need to have seperate code for differing types of observational data.

This approach (using a relational DB) also allows for using database aggregation functions to reduce the amount of data sent to the user, or that needs processing to produce a plot. For example, the current code for showing drifters on the OpenLayers map first pulls the entire track (which is roughly hourly data), and then tries to simplify the track based on the user's window. With the database approach, we can easily restrict the track to a specific timeframe, and also average out the locations based on a supplied time quantum; if a user wants to know the path of the drifters over an entire 3 year lifetime, monthly or weekly averages will be enough, and will be much faster.

## Screenshot(s)
image

## Checks
- [X] I ran unit tests.
  As the data model only, this code shouldn't even get imported yet until the next set of changes are ready. So this will have no impact on the user yet.

```sh
```


geoffholden:
> So far so good.
> To help me out with visualizing things, could you attach a table diagram (e.g. https://dbdiagram.io/home or something similar) please?
> I'll continue reviewing afterwards.

https://dbdiagram.io/d/5e70e2d14495b02c3b885c95

data/observational/orm/platform.py
```diff
@@ -0,0 +1,51 @@
+from types import MappingProxyType
+from typing import Dict, List
+import enum
+
+from data.observational import db
+
+class Platform(db.Model):
+    __tablename__ = 'platforms'
+
+    id = db.Column(db.Integer, primary_key=True)
```
htmlboss: For anyone else down the road we don't need to explicitly set `auto_increment` here since sqlalchemy sets the first non-foreign-key integer column to auto increment.

tests/test_observational_queries.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+        if 'time' in df.columns:
+            df['time'] = pd.to_datetime(df['time'])
+
+        df.to_sql(name, db.engine, if_exists='append', index=False)
+
+
+    @classmethod
+    def insertTestData(cls):
+        cls.__json_to_sql("platforms")
```
htmlboss: That's a cool trick.

PR #686, opened 2020-03-18
Updated the THREDDS URL.

dwayne-hart:
Corrected the URL link to point to the correct THREDDS instance.


dwayne-hart:
Just a change to the [...]/oceannavigator/frontend/index.html text file.

Issue #673, opened 2020-03-03
API Stability: Virtual Mooring Plot


jamesprayner:
**Describe the bug**
Requesting a Virtual Mooring plot doesn't return an image, instead it returns:
`IndexError: index 0 is out of bounds for axis 0 with size 0`

full stack:
```
Traceback (most recent call last):
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/buildadm/miniconda/3/ubuntu/16.04/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
[...]

PR #687, opened 2020-03-19
Flesh out support for timeseries data.

htmlboss:
Partially addresses #673 

tldr: virtual mooring is back (except for bottom depths)!

FVCOM will come at a later date since @geoffholden has work that depends on the fixes in here, so time is an issue. We don't have any operational FVCOM datasets currently, nor do we expect such datasets in the short-to-medium term.

**Note** timeseries for "bottom" depths won't be addressed in this PR for the above reasoning (i.e. remain broken). I believe there is some numpy axis magic possible to reduce the complexity of our bottom calculations and reduce the volume of code.

## Background

Our timeseries plots have been broken from some time so the basic idea behind these changes is to fix all that. At a lower level, I've tried to simplify the "data getter" functions in `Model.py` so callers can explicitly pass an `endtime` argument to indicate a timeseries is being requested. Before, we were doing implicit checks of the `time`/`timestamp` arguments which resulted in a vague API and code instability.

This PR will re-implement support for timeseries data for any depth level, and "All" depths. "Bottom" depths will come soon.

Maybe not the simplest way to do things, but I've opted to delete some dead code (will be commented below) and factor out some code into `Model` thus reducing duplication in the classes deriving from `Model`. This is a first pass at cleaning things up. Further cleaning can come later and is outside the scope of this PR.

I had to make some minor alterations to our resampling functions in terms of data layout. No changes were made to the actual interpolation though. Screenshots below demonstrate no change in the final result, as intended. Now that we pass slice objects for the time dimension when extracting data, this has caused some issues where the time dim size == 1. To remedy this, I've sprinkled some numpy squeezes to get things back to their expected shape (I compared the shapes of the old resampling vs what I have here and there was a difference in the shapes). Maybe not the "best" approach but the scope was rapidly exploding at that point.

## Screenshot(s)
Virtual mooring (all depths) - Mercator:
image
Virtual mooring (depth 0) - Mercator:
image
Virtual mooring (all depths) - Nemo:
image
Virtual mooring (depth 1) - Nemo:
image

Hovmoller - Nemo:
image
Hovmoller - Mercator:
image

Change comparisons

Profiles:
Before:
image
After:
image

Transects:
Before:
image
After:
image

Deep Sound Channel (calculated):
image

No behaviour change to calculated variables or map tiles:
image

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -153,7 +153,7 @@ def sspeed(depth: Union[np.ndarray, xr.Variable],
         press = press[..., np.newaxis]
 
     speed = seawater.svel(salinity, temperature, press)
-    return np.array(speed)
+    return np.squeeze(speed)
 
 
 def density(depth, latitude, temperature, salinity) -> np.ndarray:
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -80,7 +80,7 @@ def __validate_depth_lat_temp_sal(depth, latitude, temperature, salinity):
     if type(salinity) is not np.ndarray:
         salinity = np.array(salinity)
 
-    return depth, latitude, temperature, salinity
+    return depth, latitude, np.squeeze(temperature), np.squeeze(salinity)
 
 
 def __find_depth_index_of_min_value(data: np.ndarray, depth_axis=0) -> np.ndarray:
--- a/data/fvcom.py
+++ b/data/fvcom.py
@@ -9,7 +9,6 @@ import pytz
 from cachetools import TTLCache
 from pykdtree.kdtree import KDTree
 
-import data.geo as geo
 from data.calculated import CalculatedData
 from data.model import Model
 from data.netcdf_data import NetCDFData
@@ -246,24 +245,6 @@ class Fvcom(Model):
             output = output.reshape(origshape[1:])
 
         return np.squeeze(output)
-
-    def get_path(self, path, depth, time, variable, numpoints=100, times=None, return_depth=False):
-        if times is None:
-            if hasattr(time, "__len__"):
-                times = self.nc_data.timestamp_to_iso_8601(time)
-            else:
-                times = None
-        distances, times, lat, lon, bearings = \
-            geo.path_to_points(path, numpoints, times=times)
-
-        if return_depth:
-            result, dep = self.get_point(lat, lon, depth, time, variable,
-                                         return_depth=return_depth)
-            return np.array([lat, lon]), distances, times, result, dep
-        else:
-            result = self.get_point(lat, lon, depth, time, variable,
-                                    return_depth=return_depth)
-            return np.array([lat, lon]), distances, times, result
 
     def get_raw_point(self, latitude, longitude, depth, timestamp, variable):
         min_i, max_i, radius = self.__bounding_box(
--- a/data/mercator.py
+++ b/data/mercator.py
@@ -4,7 +4,6 @@ import numpy as np
 import pyresample
 from pint import UnitRegistry
 
-import data.geo as geo
 from data.calculated import CalculatedData
 from data.model import Model
 from data.nearest_grid_point import find_nearest_grid_point
@@ -81,22 +80,15 @@ class Mercator(Model):
         return np.int64(miny), np.int64(maxy), np.int64(minx), np.int64(maxx), np.amax(50000)
 
     def __resample(self, lat_in, lon_in, lat_out, lon_out, var, radius=50000):
-        if len(var.shape) == 3:
-            var = np.rollaxis(var, 0, 3)
-
+        var = np.squeeze(var)
+        
         origshape = var.shape
 
-        data = np.ma.masked_invalid(var[:])
-
-        lon_in, lat_in = pyresample.utils.check_and_wrap(lon_in, lat_in)
-
-        masked_lon_in = np.ma.array(lon_in)
-        masked_lat_in = np.ma.array(lat_in)
-
-        output_def = pyresample.geometry.SwathDefinition(
-            lons=np.ma.array(lon_out),
-            lats=np.ma.array(lat_out)
-        )
+        data, masked_lat_in, masked_lon_in, output_def = super()._make_resample_data(lat_in,
+                                                                                     lon_in,
+                                                                                     lat_out,
+                                                                                     lon_out,
+                                                                                     var)
 
         if len(data.shape) == 3:
             output = []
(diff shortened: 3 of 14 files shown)


htmlboss:
Last thing to do is Hovmoller plots

htmlboss:
Type hinting will be coming soon too.

data/fvcom.py
```diff
@@ -247,24 +247,6 @@ def weight(r):
 
         return np.squeeze(output)
 
-    def get_path(self, path, depth, time, variable, numpoints=100, times=None, return_depth=False):
```
htmlboss: I've deleted `get_path` from fvcom, nemo, and mercator and moved it into model, since the implementations were identical.

data/mercator.py
```diff
@@ -132,28 +125,17 @@ def __resample(self, lat_in, lon_in, lat_out, lon_out, var, radius=50000):
...
+            output = output.reshape((
+                origshape[0], # time
+                origshape[1], # depth
+                output.shape[0], # lat
+                output.shape[1] # lon
+            ))
 
         return np.squeeze(output)
 
-    def get_path(self, path, depth, time, variable, numpoints=100, times=None, return_depth=False):
```
htmlboss: I've deleted `get_path` from fvcom, nemo, and mercator and moved it into model, since the implementations were identical.

data/nemo.py
```diff
@@ -177,24 +173,6 @@ def __latlon_vars(self, variable):
 
         raise LookupError("Cannot find latitude & longitude variables")
 
-    def get_path(self, path, depth, time, variable, numpoints=100, times=None, return_depth=False):
```
htmlboss: I've deleted `get_path` from fvcom, nemo, and mercator and moved it into model, since the implementations were identical.

plotting/hovmoller.py
```diff
@@ -61,50 +59,24 @@ def find_depth(depth, clip_length, dataset):
...
-            if self.starttime > 0:
-                self.starttime = dataset.nc_data.timestamp_to_time_index(
-                    self.starttime)
-            if self.endtime > 0:
-                self.endtime = dataset.nc_data.timestamp_to_time_index(self.endtime)
-
-            time = time_var[slice(min(self.starttime, self.endtime), max(
-                self.starttime, self.endtime) + 1)].values
-
-            if len(self.variables) > 1:
```
htmlboss: this if-block is dead code; a relic from when we didn't have the calculation layer to produce the magnitude of two components so we did it manually. There are tons of other spots in the code that do this, but I figured since I'm editing this file already, let's just start cleaning it up.

plotting/hovmoller.py
```diff
@@ -113,39 +85,20 @@ def find_depth(depth, clip_length, dataset):
...
+            with open_dataset(compare_config, timestamp=self.compare['starttime'], endtime=self.compare['endtime'], variable=self.compare['variables']) as dataset:
                 self.compare['depth'], self.compare['depth_value'], self.compare['depth_unit'] = find_depth(
                     self.compare['depth'], len(dataset.depths) - 1, dataset)
 
-                self.fix_startend_times(
-                    dataset, self.compare['starttime'], self.compare['endtime'])
-                time = list(
-                    range(self.compare['starttime'], self.compare['endtime'] + 1))
-
-                if len(self.compare['variables']) > 1:
```
htmlboss: dead code for same reason as above.

plotting/hovmoller.py
```diff
@@ -113,39 +85,20 @@ def find_depth(depth, clip_length, dataset):
         # Load data sent from Right Map (if in compare mode)
         if self.compare:
             compare_config = DatasetConfig(self.compare['dataset'])
-            with open_dataset(compare_config, variable=self.compare['variables']) as dataset:
+            with open_dataset(compare_config, timestamp=self.compare['starttime'], endtime=self.compare['endtime'], variable=self.compare['variables']) as dataset:
```
htmlboss: This will resolve a crash when trying to open the compared dataset. No other testing done beyond this since it's outside the scope of this PR.

plotting/timeseries.py
```diff
@@ -43,25 +43,24 @@ def parse_query(self, query):
...
-
-            if len(self.variables) == 1:
-                self.variable_unit = self.dataset_config.variable[
-                    dataset.variables[self.variables[0]]
-                ].unit
-                self.variable_name = self.dataset_config.variable[
-                    dataset.variables[self.variables[0]]
-                ].name
-            else:
-                self.variable_name = self.get_vector_variable_name(dataset,
```
htmlboss: Dead code

plotting/timeseries.py
```diff
@@ -72,36 +71,34 @@ def load_data(self):
...
+            if not (set(dataset.variables[variable].dimensions) &
                     set(dataset.nc_data.depth_dimensions)):
                 self.depth = 0
 
-            times = None
             point_data = []
+            dep = []
             for p in self.points:
                 data = []
-                for v in self.variables:
```
htmlboss: No need for a loop here since we have the calculation layer to produce correct magnitude values. This plot doesn't support multiple variables on the same plot anyway.

plotting/timeseries.py
```diff
@@ -161,7 +158,7 @@ def csv(self):
                 header.append(["Depth", self.depth])
             else:
                 header.append(
-                    ["Depth", "%d%s" % (np.round(self.depths[self.depth]),
+                    ["Depth", "%.4f%s" % (self.depths,
```
htmlboss: Increase precision of depth value being saved to CSV.

plotting/timeseries.py
```diff
@@ -132,12 +127,14 @@ def load_data(self):
...
-            self.depths = depths
+            self.depths = dep
             self.depth_unit = "m"
 
+
     def csv(self):
         header = [
-            ['Dataset', self.dataset_name],
+            ['Dataset', self.dataset_config.name],
+            ['Attribution', self.dataset_config.attribution]
```
htmlboss: Crediting the original dataset is kinda important.

plotting/timeseries.py
```diff
@@ -132,12 +127,14 @@ def load_data(self):
...
             self.data = point_data
-            self.depths = depths
+            self.depths = dep
             self.depth_unit = "m"
 
+
     def csv(self):
         header = [
-            ['Dataset', self.dataset_name],
+            ['Dataset', self.dataset_config.name],
```
htmlboss: Use the user-friendly dataset name, not the internal snake-case dataset key.

plotting/timeseries.py
```diff
@@ -253,22 +250,12 @@ def csv(self):
         return super(TimeseriesPlotter, self).csv(header, columns, data)
 
     def plot(self):
-        if len(self.variables) > 1:
```
htmlboss: dead code for same reason as above.

plotting/timeseries.py
```diff
@@ -72,36 +68,34 @@ def load_data(self):
...
 
-            if not (set(dataset.variables[var].dimensions) &
+            # Override depth request if requested variable has no depth (i.e. surface only)
+            if not (set(dataset.variables[variable].dimensions) &
                     set(dataset.nc_data.depth_dimensions)):
                 self.depth = 0
 
-            times = None
             point_data = []
+            dep = []
```
jamesprayner: I think `dep` should be renamed to `depths` or something similar just for clarity

data/calculated_parser/functions.py
```diff
@@ -80,7 +80,7 @@ def __validate_depth_lat_temp_sal(depth, latitude, temperature, salinity):
     if type(salinity) is not np.ndarray:
         salinity = np.array(salinity)
 
-    return depth, latitude, temperature, salinity
+    return depth, latitude, np.squeeze(temperature), np.squeeze(salinity)
```
htmlboss: some of that squeezing I was talking about.

PR #688, opened 2020-03-19
Erddap access

douglatornell:
## Background
Add the ability to load datasets from ERDDAP servers via OPeNDAP dataset URLs. 

In particular, enable datasets to be loaded from the SalishSeaCast ERDDAP server. Those datasets require the extra optional feature of merging in a dataset that maps the model x-y grid indices to longitude-latitude.

It was the obvious path after having been immersed in the `data` modules doing the `open_dataset()` refactoring for PR#641,

This adds an optional `geo_ref` item to the datasetconfig shcema. I will update the wiki to add it tomorrow.

Presently there is only one of the SalishSeaCast tracer datasets (sea surface height) included in the dataset config. I can add more to this PR if I get them pushed before it is merged, but merging doesn't need to wait on them. They can just as easily be separate PRs.

## Screenshot(s)
image

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/routes/api_v1_0.py
+++ b/routes/api_v1_0.py
@@ -766,14 +766,17 @@ def timestamps():
         raise APIError("Please specify a variable via ?variable=variable_name")
     variable = args.get("variable")
 
-    vals = []
-    with SQLiteDatabase(config.url) as db:
-        if variable in config.calculated_variables:
-            data_vars = get_data_vars_from_equation(config.calculated_variables[variable]['equation'],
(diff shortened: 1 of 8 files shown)


routes/api_v1_0.py
```diff
@@ -1,7 +1,6 @@
 import base64
 import datetime
 import gzip
-import hashlib
```
htmlboss: Nice :+1:

PR #695, opened 2020-03-24
Fix typo

htmlboss:
## Background
Fix a typo

It's the only way

## Checks
- [x] I ran unit tests.

```sh
```


htmlboss:
ty will merge after freeze is lifted

PR #696, opened 2020-03-24
Tweak open_dataset to ingest multiple variables at once

htmlboss:
Resolves T/S plot crashes.

When the sql back-end was implemented, `open_dataset` didn't consider multiple variables in the `variable` kwarg. This resulted in T/S plots failing for datasets that had variables split across multiple files. This PR fixes that.

## Background

* Also updated a netcdf fixture to re-enable some API tests.

Path of least resistance. This is actually finishing a part of the sqlite refactor I started back in Aug 2019. Forgot to finish it when Fall came around.

More tests have been enabled!

## Screenshot(s)
image

## Checks
- [x] I ran unit tests.

```sh
```


data/netcdf_data.py
```diff
@@ -757,34 +757,46 @@ def get_nc_file_list(self, datasetconfig: DatasetConfig, **kwargs: dict) -> Unio
...
+            if not isinstance(variable, set):
+                variable = set(variable)
 
-            calculated_variables = datasetconfig.calculated_variables
-            if variable[0] in calculated_variables:
-                equation = calculated_variables[variable[0]]['equation']
-
-                variable = data.utils.get_data_vars_from_equation(
-                    equation, [v.key for v in db.get_data_variables()])
+            raise
```
htmlboss: temporary raise for debugging. will be deleted before requesting review

data/netcdf_data.py
```diff
@@ -757,34 +757,46 @@ def get_nc_file_list(self, datasetconfig: DatasetConfig, **kwargs: dict) -> Unio
...
-                variable = [variable]
+            if not isinstance(variable, set):
+                variable = set(variable)
 
-            calculated_variables = datasetconfig.calculated_variables
-            if variable[0] in calculated_variables:
-                equation = calculated_variables[variable[0]]['equation']
-
-                variable = data.utils.get_data_vars_from_equation(
-                    equation, [v.key for v in db.get_data_variables()])
```
douglatornell: I stumbled across this issue a few days ago when I noticed that CTD profiles and T/S diagrams weren't working. Sorry, I didn't mention it, or provide a patch, but here is the solution I came up with:
```python
            if not isinstance(variable, list):
                variable = set([variable])
            else:
                variable = set(variable)

            calculated_variables = datasetconfig.calculated_variables
            for var in variable:
                if var in calculated_variables:
                    equation = calculated_variables[var]['equation']
                    equation_vars = data.utils.get_data_vars_from_equation(
                            equation, [v.key for v in db.get_data_variables()])
                    variable.union(set(equation_vars))
            variable = list(variable)
```
More casting back and forth between `list` and `set` than I would like - perhaps some can be polished. Using `set` to avoid duplication of vars from eqns, IIRC.

Grist for the mill...

htmlboss: No worries! It's a reassurance for me that my solution is similar in thinking to yours :1st_place_medal:
To avoid the `variable.union(set(equation_vars))` cast,  I used a `variable.update(equation_vars)` which grabs stuff from other iterables.

data/netcdf_data.py
```diff
@@ -757,32 +757,42 @@ def get_nc_file_list(self, datasetconfig: DatasetConfig, **kwargs: dict) -> Unio
...
+            variables_to_load = self.__get_variables_to_load(db, variable, calculated_variables)
 
             try:
                 timestamp = self.__get_requested_timestamps(
-                    db, variable[0], kwargs['timestamp'], kwargs.get('endtime'), kwargs.get('nearest_timestamp', False))
+                    db, variables_to_load[0], kwargs['timestamp'], kwargs.get('endtime'), kwargs.get('nearest_timestamp', False))
```
htmlboss: the timestamp and endtime kwargs are the same for all variables when loading, so just grab the first variable.

data/netcdf_data.py
```diff
@@ -757,32 +757,42 @@ def get_nc_file_list(self, datasetconfig: DatasetConfig, **kwargs: dict) -> Unio
...
             if not file_list:
                 raise RuntimeError("NetCDF file list is empty.")
 
             self._nc_files = file_list
+    
+    def __get_variables_to_load(self, db: SQLiteDatabase, variable: set,
+                                    calculated_variables: dict) -> List[str]:
+        
+        calc_var_keys = set(calculated_variables)
+        variables_to_load = variable.difference(calc_var_keys)
```
htmlboss: I could do `variable - calc_var_keys` here but I think `.difference` is more clear.

data/sqlite_database.py
```diff
@@ -38,7 +38,7 @@ def __exit__(self, exc_type, exc_value, traceback):
     def __flatten_list(self, some_list: list) -> list:
         return list(itertools.chain(*some_list))
 
-    def get_netcdf_files(self, timestamp: list, variable: list) -> List[str]:
+    def get_netcdf_files(self, timestamp: List[int], variable: List[str]) -> List[str]:
```
htmlboss: drive-by type hint improvement.

jamesprayner:
Have you tried running the tests with the new subset file AFTER clearing your navigator cache? Just want to make sure they all still work.

htmlboss:
> Have you tried running the tests with the new subset file AFTER clearing your navigator cache? Just want to make sure they all still work.

Good catch. Just tried it and no additional problems.

PR #704, opened 2020-03-30
Ditch Class4 dependency on THREDDS & FTP

htmlboss:
## Background
Up to now, our class4 architecture depended on THREDDS + FTP to serve our Class4 files. The use case was to simply create lists of files and to be able to access them...kinda how a basic filesystem works right? We weren't using the full functionality of FTP or THREDDS to warrant them being kept as dependencies for this part of the code base.

So this PR simply tweaks the class4 pickle generation code primarily (and other spots) to `glob` a given `CLASS4_PATH` in the oceannavigator.cfg file. No more network dependency. This resulted in the pickle file generation time on Dory to be reduced significantly (by a factor of 5 at least).

I also switch to using xArray to make things in line with our other dataset handling code.

Didn't want to end up re-writing the class4 architecture at this time, so this was the cleanest way I could think of to make what we have usable again.

The plotting window sort-of works in the sense the class4 forecasts and model list endpoints are still broken. They will be fixed in another PR.

It'll take a few seconds for the dots to appear on the map, but they do so reliably now. No more time outs.

Code is simpler too.

## Screenshot(s)

Main map:
image

Plot window:
image

## Checks
- [x] I ran unit tests.

```sh
```


oceannavigator/oceannavigator.cfg
```diff
@@ -6,9 +6,8 @@ OVERLAY_KML_DIR = "./kml"
 DRIFTER_AGG_URL = "http://10.0.3.56:8080/thredds/dodsC/Drifter/aggregated.ncml"
 DRIFTER_URL = "http://10.0.3.56:8080/thredds/dodsC/Drifter/%s.nc"
 DRIFTER_CATALOG_URL = "http://10.0.3.56:8080/thredds/catalog/Drifter/catalog.xml"
-CLASS4_CATALOG_URL = "http://10.0.3.56:8080/thredds/catalog/Class4/catalog.xml"
-CLASS4_URL = "http://10.0.3.56:8080/thredds/dodsC/Class4/%s/%s.nc"
-CLASS4_FTP = "10.0.3.9"
+CLASS4_FNAME_PATTERN = "/data/class4/%s/%s.nc"
```
htmlboss: Did a find+replace for this. A bit funky but as I said I don't want to radically change the overall architecture here.

scripts/generate_class4_list.py
```diff
@@ -14,33 +14,31 @@
...
+def list_class4_files(class4_path):
     files = []
     result = []
-    for i in range(2011, thisYear + 1):
-        fullList = ftpSite.nlst("/class4/" + str(i))
-        for filename in fullList:
-            if ("_GIOPS_" in filename) and ("profile.nc" in filename):
-                files.append(filename[13:-3])
+    
+    for f in glob.iglob(f"{class4_path}/**/*.nc", recursive=True):
```
htmlboss: `glob` is more pythonic apparently.

PR #705, opened 2020-03-30
Observational API

geoffholden:
## Background
This adds the new Observational API JSON requests.
The new requests are:
/api/v1.0/observation/datatypes.json
- Gets the list of observational data types in the database

/api/v1.0/observation/meta_values/<string:platform_types>/<string:key>.json
- Gets all the metadata values for a given key and list of platform types

/api/v1.0/observation/tracktime/<string:platform_id>.json
- Gets the start and end times for a platform's track

/api/v1.0/observation/track/<string:query>.json
- Gets the track data for the query

/api/v1.0/observation/point/<string:query>.json
- Gets the point data for a query

/api/v1.0/observation/meta.json
- Gets all the metadata for 

/api/v1.0/observation/variables/<string:query>.json
- Gets the list of variables for an individual platform or station

I would have liked to not have all the <string:query> arguments and use a proper , but without changing the ComboBox JSX component (which I think nobody wants to touch with a 10' pole at this point), the URL needs be constructed by hand. I don't believe that building query parameters by hand is the right call (especially when jQuery's .ajax() function can take a 'data' argument), but I'm willing to be convinced to switch that one around.

The old queries are still present as well. I'll remove them when the new UI PR.

## Checks
- [X] I ran unit tests.

```sh
```


routes/api_v1_0.py
```diff
@@ -915,6 +921,382 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
...
+    Queries the min and max times for the track
+
+    **Used in TrackWindow**
+    """
+    max_age = 86400
+    platform = DB.session.query(Platform).get(platform_id)
+    data = DB.session.query(
+        DB.func.min(Station.time),
+        DB.func.max(Station.time),
+    ).filter(Station.platform == platform).one()
```
htmlboss: https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.one

routes/api_v1_0.py
```diff
@@ -915,6 +921,382 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
...
+    """
+    max_age = 86400
+    data = ob_queries.get_meta_values(
+        DB.session, platform_types.split(','), key
+    )
+    resp = jsonify(data)
+    resp.cache_control.max_age = max_age
+    return resp
+
+@bp_v1_0.route('/api/v1.0/observation/tracktime/<string:platform_id>.json')
```
htmlboss: Could we increase specificity by using `tracktimerange` or something similar? Looking to indicate from this declaration that the URL handles a time range.

geoffholden: Done.

htmlboss:
> which I think nobody wants to touch with a 10' pole at this point

I think you think correctly :wink: 

I agree the queries are a bit janky, but I can see the method to the madness. Once I see the front-end changes, I might be able to suggest a thing or two.

routes/api_v1_0.py
```diff
@@ -915,6 +921,382 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
...
+
+        params['platforms'] = platforms
+
+    coordinates = ob_queries.get_platform_tracks(
+        DB.session,
+        query_dict.get("quantum", "day"),
+        **params
+    )
+
+    if len(coordinates) == 0:
```
htmlboss: I think this could be shortened to `if not coordinates`...implicitly it does the length check.

routes/api_v1_0.py
```diff
@@ -915,6 +921,382 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
...
+        coordinates = [[0, 0, 0, 0]]
+
+    df = pd.DataFrame(np.array(coordinates), columns=['id', 'type', 'lon', 'lat'])
+    df['id'] = df.id.astype(int)
+
+    if len(coordinates) > 1:
```
htmlboss: Same thing here `if coordinates`...implicit check.

geoffholden: I just removed one check, the other needs to stay as len(coordinates) because it's checking for at least 2 points--no such thing as a track with only one point.

geoffholden:
> Any chance for some api tests? I think mocking all of the DB return values would make it pretty straight forward since they're already being tested elsewhere. For now we're only looking for 200 status codes and expected exception raises but extra stuff is good too.

Yeah, I'll see what I can do with that tonight/tomorrow.

Issue #702, opened 2020-03-27
Replace seawater package with gsw


douglatornell:
**Describe the bug**
The `seawater` (https://pythonhosted.org/seawater/) is considered obsolete because it is based on the the old EOS-80 oceanographic tables and standards. EOS-80 has been replaced by the [TEOS-10](http://www.teos-10.org/) formulation that derives all seawater properties in a thermodynamically consistent manner. TEOS-10 is implemented in the `gsw` package (https://github.com/TEOS-10/GSW-Python) which is installable from `conda-forge`.

**To Reproduce**
Try to convert conservative temperature to potential temperature. 

**Expected behavior**
Correct calculations involving seawater properties.

**Screenshots**
N/A

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
All
 
**Additional context**
This is unlikely to be an easily resolved issue because it is scientifically complicated. There may not be a simple 1:1 mapping of functions between `seawater` and `gsw` for all of the use-cases in Navigator. Having some form of verification plan for the conversion would be good because (as we all know) it is easy to feed number of equations and get numbers out, but are they correct numbers?

PR #706, opened 2020-03-30
Observational Data Importers

geoffholden:
## Background
I'm not sure if you'll want these as part of the repo or not--probably so for 'sample' ones, but possibly not going forward if you get the data providers to manage their own importers.

These have been somewhat optimized for import speed (they're still pretty slow) by going around the ORM a little by using things like bulk_save_objects that doesn't follow relationships. This cuts down on a lot of DB transactions and speeds things up a lot at the expense of a little readability.

For anybody writing a new importer, my first suggestion would be to switch echo to True in init_db so you'll see every SQL statement printed to screen. With that set, you can use an in-memory sqlite database (URI sqlite:///:memory:) so that you can mess with things a fair bit without touching the proper DB. Even after that, you can test on file-backed sqlite locally.

## Checks
- [ ] I ran unit tests.

```sh
```

### DIFF
--- a/scripts/data_importers/argo.py
+++ b/scripts/data_importers/argo.py
@@ -42,9 +42,9 @@ def main(uri: str, filename: str):
     else:
         filenames = [filename]
 
-    for filename in filenames:
-        print(filename)
-        with xr.open_dataset(filename) as ds:
+    for fname in filenames:
+        print(fname)
+        with xr.open_dataset(fname) as ds:
             times = pd.to_datetime(ds.JULD.values)
 
             for f in META_FIELDS:
--- a/scripts/data_importers/argo.py
+++ b/scripts/data_importers/argo.py
@@ -1,11 +1,9 @@
 #!env python
 
-import numpy as np
 import pandas as pd
 import xarray as xr
 import defopt
 import seawater
-import datetime
 import os
 import glob
 import data.observational
--- a/scripts/data_importers/argo.py
+++ b/scripts/data_importers/argo.py
@@ -1,4 +1,4 @@
-#!env python
+#!/usr/bin/env python
 
 import pandas as pd
 import xarray as xr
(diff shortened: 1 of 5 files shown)


scripts/data_importers/argo.py
```diff
@@ -0,0 +1,131 @@
+#!env python
+
+import pandas as pd
+import xarray as xr
+import defopt
+import seawater
```
htmlboss: FYI #702 No need to do anything about this now tho.

geoffholden:
> Looks good. I can see you faced a similar issue w.r.t. file processing speed like I did with the index tool. I think if we shove the observational data on Dorys ssds, it'll cut things down a fair bit.

I don't think that'll make much of a difference in this case, as the slowdown was in database transactions. Particularly because it's normalized so we need the id of table a to create the items in table b, etc. Now, I was using a remote connection to MySQL/MariaDB, so having it on the same host that the importing is running should speed things along a bit too.

One other note, these scripts use the defopt package that I don't believe is in the standard environment for the navigator (but you can get it via pip, I don't think it's on conda-forge). It is a convenient package though, especially if you find setting up argument parsers a PITA.

scripts/data_importers/argo.py
```diff
@@ -0,0 +1,131 @@
...
+    :param str uri: Database URI
+    :param str filename: Argo NetCDF Filename, or directory of files
+    """
+    data.observational.init_db(uri, echo=False)
+    data.observational.create_tables()
+
+    session = data.observational.db.session
+
+    if os.path.isdir(filename):
+        filenames = sorted(glob.glob(os.path.join(filename, "*.nc")))
```
htmlboss, line 39: I'm curious about the sorting here...

htmlboss:
@dwayne-hart also updated the environment repo for `defopts`.

scripts/data_importers/argo.py
geoffholden: Yeah, that doesn't _need_ to be sorted. I had that in there because I was getting disconnection issues and I wanted to be able to pick up where I left off, but a default glob isn't necessarily sorted.

PR #707, opened 2020-03-31
Observational UI

### LINKED ISSUE #168: Add ARGOS
There will need to be some thought given to how this will work when plotting the graphs, since there is a depth portion to them as well.

### LINKED ISSUE #169: Add Gliders
There will need to be some thought given to how this will work when plotting the graphs, since there is a depth portion to them as well.

geoffholden:
Resolves #169 
Resolves #168
## Background
This has the Javascript UI updates and the Python plotting functions for the new observational data.

What was once called DrifterWindow and DrifterPlotter have now been renamed to TrackWindow and TrackPlotter, as that's more general and covers animal tracks, gliders, argo floats, etc.
I removed the Observations menu item under the Point menu, and removed the Drifter menu, replacing all of that with a new Observations menu. That menu now has the options to select which observations to show, and also has a shortcut to show the current active drifters (active in the last 30 days, with daily average locations plotted)

For plotting tracks, it used to use time as the x-axis, it's now using distance instead. It seems like that's the more common way to visualize this kind of data (and unfortunately it doesn't work well as a secondary axis, since they're not necessarily related to each other).

## Screenshot(s)
Screen Shot 2020-03-31 at 12 26 26 PM
Screen Shot 2020-03-31 at 12 26 06 PM
Screen Shot 2020-03-31 at 12 25 42 PM
Screen Shot 2020-03-31 at 12 24 46 PM

## Checks
- [X] I ran unit tests.

```sh
```


htmlboss:
Note `https://pypi.org/project/visvalingamwyatt/` needs to be installed via `pip`

plotting/observation.py
```diff
@@ -2,21 +2,20 @@
...
 from flask_babel import format_datetime, gettext
-from netCDF4 import Dataset
 
 from data import open_dataset
+from data.utils import datetime_to_timestamp
 from plotting.point import PointPlotter
 from plotting.utils import mathtext
 from utils.errors import ClientError
 
+from data.observational import db, Station, Station, Sample, DataType
```
htmlboss, line 18: Double import for `Station`

plotting/track.py
```diff
@@ -0,0 +1,441 @@
...
+from data.observational import db, Platform, DataType, Station, Sample
+from data.observational.queries import get_platform_variable_track
+from plotting.plotter import Plotter
+
+
+class TrackPlotter(Plotter):
+
+    def __init__(self, dataset_name: str, query: str, **kwargs):
+        self.plottype: str = "track"
+        super(TrackPlotter, self).__init__(dataset_name, query, **kwargs)
```
htmlboss, line 28: Python 3 doesn't need the arguments in `super` anymore. They're implicit. `super().__init__(dataset_name, query, **kwargs)`. This applies to the other instances of this below.

htmlboss:
I'll be sitting down with @dwayne-hart to ensure we're good to go on the back-end side before merging in.

PR #708, opened 2020-04-01
Fix data.class4.list_class4_models crash

htmlboss:
## Background
#704 forgot to remove the FTP code in `data.class4.list_class4_models` so a crash was occurring. This PR just re-writes that function to use the same `glob` pattern established in #704.

I renamed `tests.test_api_v_1_0.test_class4_query_endpoint` to `tests.test_api_v_1_0.test_class4_models_endpoint`. It has also been enabled.

Simplest way.

I still don't like how we figure out which class4 file to open but that can be re-done another time.

## Screenshot(s)
Alternate models can be selected again:
image

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/class4.py
+++ b/data/class4.py
@@ -4,6 +4,7 @@ import glob
 import os
 import pickle as pickle
 import time
+from typing import List
 
 import cftime
 import numpy as np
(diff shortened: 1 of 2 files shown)


data/class4.py
```diff
@@ -226,22 +227,28 @@ def list_class4_forecasts(class4_id):
...
-    ftpSite = FTP(current_app.config['CLASS4_FTP'])
-    ftpSite.login("anonymous", "anonymous")
-    thisYear = datetime.datetime.now().year
-    files = []
-    for i in range(2011, thisYear + 1):
-        fullList = ftpSite.nlst("/class4/" + str(i))
-        for filename in fullList:
-            if "profile.nc" in filename:
-                files.append(filename[13:-3])
+def list_class4_models(class4_id: str) -> List[dict]:
```
htmlboss, line 230: drive-by type hints and docstring

tests/test_api_v_1_0.py
```diff
@@ -209,8 +209,7 @@ def test_data_endpoint(self, patch_get_data_vars, patch_get_dataset_config):
         self.assertEqual(res.status_code, 200)
 
 
-    @unittest.skip("Skipping api/class4.. needs re-write")
-    def test_class4_query_endpoint(self):
+    def test_class4_models_endpoint(self):
```
htmlboss, line 212: yay more test coverage

PR #709, opened 2020-04-02
Add SalishSeaCast currents dataset config, colour maps & dataset handling

douglatornell:
## Background
This adds the 3D u and v velocity component datasets from SalishSeaCast and the speed of the seawater current calculated from them. 

It was the logical thing to do.

The design concept for this is described in the Slack post at https://ocean-navigator.slack.com/archives/CS85QEFLJ/p1583704247069300

Unfortunately, it was not possible to extend `functions.magnitude()` as described in the design concept, so a new `functions.unstaggered_speed()` was implemented.

Although I described the use of a list of URLs in the dataset config as a "smelly hack" in the 30-Mar-2020 tech meeting, I now think that it is a feature that has value. Because of lazy-loading, the variables from a list of ERDDAP/OPeNDAP URLs that point at datasets with compatible coordinates can be merged into a single `NetCDFData` instance. That makes possible the SalishSeaCast currents dataset that is the subject of this PR. It could also be used to collapse the SalishSeaCast currents, biology, and salinity & temperature datasets into a single 3D dataset. Other 3D tracer variables (density, PAR, carbon model variables) that are in the process of being added to the SalishSeaCast ERDDAP server could then simply be added to that dataset. That would avoid a proliferation of SalishSeaCast datasets in Navigator.

## Screenshot(s)
image

## Checks
- [x] I ran unit tests.

```sh
```


data/calculated_parser/functions.py
```diff
@@ -56,6 +56,40 @@ def magnitude(a, b):
...
+    v_vel: ndarray
+    x_dim_index: float
+    y_dim_index: float
+
+    Returns:
+        ndarray -- speed of current
+    """
+    # The vel(t, z, y, x) arrays have been sliced to vel(t, y, x) by the time they get here,
+    # so we subtract 1 from the dimension index values passed in from the datasetconfig.json
+    # equation.
```
htmlboss, line 85: Is this also true for transect plots? That's where we grab all depth levels and plot them against the drawn lines distance.
image

douglatornell: Oh! Nice catch! I forgot to test Points, Lines, and Areas, but I just did so now, and they didn't error out. I will do a debugging trace to figure out how the dimension reduction is happening for the Lines case, and revise that comment to make it more general wrt to dimension reduction.

htmlboss: Awesome. After that's done, I'll approve and merge.

douglatornell: Arrgh! The velocity component arrays are `(t, depth, y, x)` when they get to `unstagger_speed()` for a transect. So, it is not doing the right thing at all: shifting `u` on the `gridY` axis, and `v` on the `depth` axis. :sob: 

Two options:
1. Assume that y and x are always the last 2 dimensions, and in that order, and pick them out of the `dims` attr using -2 and -1 indices.
2. Make `unstagger_speed()` specific to SalishSeaCast by hard-coding the dimension names as `gridX` and `gridY`. In that case, does the function name need to change to reflect its specificity?

htmlboss: I think Option 1 is good. That's the layout I've been trying to implement.

htmlboss: I would have a quick comment present saying the expected shape.

douglatornell: It makes sense for us to push for `(t, depth, y, x)` or `(t, depth, lon, lat)` because that is the "standard" that NEMO uses, and NEMO is that "standard" for CONCEPTS, and CONCEPTS and Navigator are pretty closely linked...

PR #724, opened 2020-04-22
Quick and dirty clean up of R script generation.

htmlboss:
Does not check if generated R script works though. A quick look-over shows it *should* work.

## Background
Execute a blind string substitution into the R script template so the user can download an R script.

Side note: I created an empty `__init__.py` file to resolve some import errors when running unit tests on certain environment setups.

Resulting file:
```r
library(jsonlite)


query <- '{"area":[{"innerrings":[],
"name":"",
"polygons":[[[51.289405902716794,-45.9228515625],[51.34433866059925,-53.129882812499986],[46.25584681848031,-53.48144531249999],[45.82879925192134,-46.18652343749999],[51.289405902716794,-45.9228515625]]]}],
"bathymetry":1,
"colormap":"default",
"contour":{"colormap":"default",
"hatch":0,
"legend":1,
"levels":"auto",
"variable":"none"},
"dataset":"glorys4_climatology",
"depth":0,
"interp":"gaussian",
"neighbours":10,
"projection":"EPSG:3857",
"quantum":"month",
"quiver":{"colormap":"default",
"magnitude":"length",
"variable":"none"},
"radius":25,
"scale":"-5,30,auto",
"showarea":1,
"time":411356034,
"type":"map",
"variable":"votemper"}'

dpi <- 144

fname <- "script_template_" + toString(query["dataset"]) + "_" + toString(query["variable"]) + ".png"

gsub("[\r\n]", "", query)
base_url <- "http://navigator.oceansdata.ca/api/v1.0/plot/?"
url <- base_url + URLencode(query) + "&dpi=" + toString(dpi)
print(url)


print("Downloading file and exiting...")
download.file(url, fname, wget)
```

Only way.

* R script generation will crash for any request besides `PLOT`.
* Again, this does NOT guarantee the generated R script actually works. 
* There is some massive code duplication in `scriptGenerator.py` which needs to be cleaned up down the road.

## Screenshot(s)
n/a

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/plotting/scriptGenerator.py
+++ b/plotting/scriptGenerator.py
@@ -1,31 +1,7 @@
-import json
-from io import BytesIO, StringIO
-import re
-import base64
-import pytz
-import hashlib
-from data import open_dataset
-from oceannavigator import DatasetConfig
+from io import BytesIO
 
-def time_query_conversion(dataset, index):
-    """
-    API Format: /api/timestamps/?dataset=' '
 
-    dataset : Dataset to extract data - Can be found using /api/datasets
-
-    Finds all data timestamps available for a specific dataset
-    """
-
-    
-    config = DatasetConfig(dataset)
-    with open_dataset(config) as ds:
-        try:
-            date = ds.nc_data.timestamps[index]
-            return date.replace(tzinfo=pytz.UTC).isoformat()
-        except IndexError:
-            return ClientError("Timestamp does not exist")
-
-def generatePython(url, scriptType):
+def generatePython(url, scriptType: str) -> BytesIO:
     if "class4id" in url:
         var = "class4id"
     elif "drifter" in url:
@@ -43,7 +19,7 @@ def generatePython(url, scriptType):
     if "null"  in query:
         query = query.replace("null", "None")
     
-    with open("plotting/templates/python" + scriptType + "template.txt", 'r') as f:
+    with open(f"plotting/templates/python{scriptType}template.txt", 'r') as f:
         template = str(f.read())
 
         if scriptType == "PLOT":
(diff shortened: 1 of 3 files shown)


plotting/scriptGenerator.py
```diff
@@ -1,31 +1,7 @@
...
-from io import BytesIO, StringIO
-import re
-import base64
-import pytz
-import hashlib
-from data import open_dataset
-from oceannavigator import DatasetConfig
+from io import BytesIO
 
-def time_query_conversion(dataset, index):
```
htmlboss, line 10: Dead code

plotting/scriptGenerator.py
```diff
@@ -56,109 +32,34 @@ def generatePython(url, scriptType):
...
-        fileExtension = ".csv"
-    if 'output_format' in url:
-        netcdf = True
-        fileExtension = ".nc"
-        data_type = "subset"
-    #setup file
-    script = StringIO()
-    
-    #FILE CONTENTS ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    script.write("#Assigns Base Url\n\n")
```
htmlboss, line 85: Scrap all of this in favour of templating.

plotting/templates/pythonPLOTtemplate.txt
```diff
@@ -1,10 +1,12 @@
+#!/usr/bin/env python
+
 from urllib.request import urlopen
 from urllib.parse import urlencode
 from contextlib import closing
 try:
    from PIL import Image
 except:
-   print("If you are on a Windows machine, please install PIL (imaging library) using 'python -m pip install Pillow' in the Anaconda Prompt")
+   print("Please install PIL (imaging library) using 'python -m pip install Pillow'.")
```
htmlboss, line 9: Drive-by shebang and f-string.

htmlboss:
Just fixed syntax error in R template.

PR #725, opened 2020-04-22
Add warning banner

htmlboss:
## Background
Draw users' attention to new feature and also give heads up of where to log issues.

Only way

* Front-end will need a re-build as usual.

## Screenshot(s)
image

## Checks
- [x] I ran unit tests.

```sh
```


oceannavigator/frontend/src/components/WarningBar.jsx
```diff
@@ -24,20 +23,13 @@ export default class WarningBar extends React.PureComponent {
...
         </Alert>
       );
     }
 
     return(<div></div>);
   }
 }
-
-//***********************************************************************
-WarningBar.propTypes = {
```
htmlboss, line 40: Old unneeded junk.

PR #726, opened 2020-04-23
Implement Deep Sound Channel Bottom calculated variable

htmlboss:
## Background
Enabled for all GIOPS and RIOPS datasets (except for those with "Historical" in their name).

Deep sound channel bottom definition:  Find and return the deep sound channel bottom (the second depth where the speed of sound is equal to the speed of sound at the sonic layer depth).

Most efficient way in terms of CPU and memory.

100% vectorized code keeps performance impact to a minimum.

* The code is not very intuitive but it's the fastest that it can be so take time to understand it.

## Screenshot(s)
Lat/Lon grid:
image

P/S grid:
image

## Checks
- [x] I ran unit tests.

```sh
```

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -245,6 +245,23 @@ def tempgradient(depth, latitude, temperature, salinity) -> np.ndarray:
     return np.array(tempgradient)
 
 
+def __soniclayerdepth_from_sound_speed(soundspeed: np.ndarray, depth: np.ndarray) -> np.ndarray:
+
+    min_indices = __find_depth_index_of_min_value(soundspeed)
+
+    # Mask out values below deep sound channel
+    mask = min_indices.ravel()[..., np.newaxis] < np.arange(
+        soundspeed.shape[0])
+    mask = mask.T.reshape(soundspeed.shape)
+
+    soundspeed[mask] = np.nan
+
+    # Find sonic layer depth indices
+    max_indices = __find_depth_index_of_max_value(soundspeed)
+
+    return depth[max_indices]
+
+
 def soniclayerdepth(depth, latitude, temperature, salinity) -> np.ndarray:
     """
     Find and return the depth of the maximum value of the speed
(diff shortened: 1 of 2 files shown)


htmlboss:
Ignore the codefactor todo complaint

data/calculated_parser/functions.py
```diff
@@ -301,6 +322,57 @@ def deepsoundchannel(depth, latitude, temperature, salinity) -> np.ndarray:
...
+        max_indices[np.newaxis, :], # pad to equate number of dims to sound_speed
+        0 # apply along depth axis
+    )
+
+    # Flip the mask since we actually want to examine the values BELOW the sonic
+    # layer depth.
+    sound_speed.mask = ~sound_speed.mask
+
+    # Find Deep Sound Channel Bottom
+    dscb_values = np.isclose(sound_speed, sound_speed_values_at_sonic_layer_depth)
```
htmlboss, line 362: https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html

`rtol` default is 1e-05.

data/calculated_parser/functions.py
```diff
@@ -301,6 +322,57 @@ def deepsoundchannel(depth, latitude, temperature, salinity) -> np.ndarray:
...
+
+    # Now use the boolean array held in dscb_values as a mask over sound_speed.
+    # True indicates a close value. Since we want the "close" values to be visible
+    # through the mask, flip the result of np.isclose.
+    sound_speed.mask = ~dscb_values.data
```
htmlboss, line 367: image

data/calculated_parser/functions.py
nsoontie: Thanks for taking the time to vectorize!! :)

I'm not sure about using `np.isclose()` because the sound speed profile is defined at discrete points and might not ever get super close to the speed we are looking for but it might cross through that point. For example,

Depth  (m) |  Speed of Sound  (m/s) |
-------------|--------------------------|
0  | 1000 |
10 | 1010 |
20 | 1000 |
30 | 1006 |
40 | 1009 | 
50 | 1012 |
60 | 1015 |

The SLD is 10 m and the speed of sound at that point is 1010 m/s but later in the profile the speed of sound never gets "really" close to 1010 m/s.

The sound profile crosses through 1010 m/s between 40 and 50 m. 



How do you feel about interpolating using the points around where the crossing occurs?

htmlboss: it seemed too easy now that I think about it LOL I need to think about this.

NoahGallant-MUN:
I could be wrong, but I know from when I implemented deep sound channel bottom, it wasn't as spotty as in those images. Mine had larger 'clumps' where there is a sound channel bottom.
I would just verify it if you haven't already.

:)

NoahGallant-MUN:
I'm not sure if you're using my unit tests, but some might be helpful. When making them I tried to think of all the different shapes and characteristics that we might see in a sound speed profile, and then manually calculated all the different variables to check.

PR #734, opened 2020-05-29
Altering the way the certain config files are staged

dwayne-hart:
Altering the way the oceannavigator/{datasetconfig,json,oceannavigator.cfg} files are staged depending on location. At DAL or outside of that network.


dwayne-hart:
This PR is ready. I tested to ensure that my local Ocean-Data-Map-Project repository updated correctly when I performed the recursive submodule update. The symbolic links to [...]/Ocean-Data-Map-Project/oceannavigator/{datasetconfig.json,oceannavigator.cfg} pointed to the correct version of the datasets. The default  behaviour is to be production ready.

Issue #735, opened 2020-06-16
Velocity arrows show incorrect direction


nsoontie:
**Describe the bug**
In Area plots, when plotting Water Velocity arrows, the direction of the arrows is incorrect. 

**To Reproduce**
Area plot -> Water Velocity -> Add Water velocity arrows
[...]

PR #738, opened 2020-07-02
Velocity arrow fix - explicitly defined vector components

nsoontie:
## Background
This pull request addresses #735 where velocity vectors are not displaying correctly. I have explicitly defined the eastward and northward components of a vector in the datasetconfig.json and modified the code to read those vector variables rather than guessing.

While this approach adds complexity to the datasetconfig.json, it also eliminates any guesswork in identifying the vector components. This way we can be certain that correct components are being used in the quiver plots.

There is a parallel [pull request](https://github.com/DFO-Ocean-Navigator/configurations/pull/2) in the configurations repo

I did not modify any code related to quivers in the stick plots/hovmoller because that functionality is disabled at the moment.

## Screenshot(s)
Old
old

New
new

## Checks
- [ yes] I ran unit tests. 
I also added a test for the new changes.
- [ yes] I've tested the relevant changes from a user POV.


nsoontie:
I corrected the code factor issue and pulled changes from master. All unit tests passed and the website works as expected.

PR #741, opened 2020-07-21
Update to the information available to end users

dwayne-hart:
Adding in unique dataset help files which cut down on the help sections stored in the datasetconfig.json file.


dwayne-hart:
Implemented requested changes.

nsoontie:
Looks great Dwayne! I suppose there will be a parallel pull request in the configs repo?

dwayne-hart:
Hi Nancy,

There will be an update to both the main and off-site branches. I just want
to ensure that the naming convention is agreed upon to minimize the number
of commits to git. But I'm open to suggestions.

Best,
Dwayne

nsoontie:
Sounds good. Let's wait for the other changes.

PR #744, opened 2020-07-21
Add more unit tests

htmlboss:
## Background

Adds more unit tests. We're now up to 183 from 177 tests.

Only way

Nope

## Screenshot(s)
n/a

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


htmlboss:
ignore codefactor complaining about the todo and newline

PR #745, opened 2020-07-24
Fix crash when requesting a CSV for maps

htmlboss:
## Background

Fixed a crash when requesting a CSV file from the map window.

THe issue was wrong indent level lol. I broke that about 3 months ago it seems like.

Also added a new test to hit the CSV code so this shouldn't regress.

Only way.

Nope

## Screenshot(s)

Resulting CSV:
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


plotting/utils.py
```diff
@@ -18,6 +18,8 @@ def get_filename(plot_type, dataset_name, extension):
 
 
 def get_mimetype(filetype: str):
+    filetype = filetype.lower()
```
htmlboss, line 21: I realized people could send all-caps file types through the API and there is no conversion, which could break stuff. So I'll stick it to lower case here.

htmlboss:
cc @VanessaSutton-Pande

PR #747, opened 2020-08-07
Mask out surface values for some acoustic variables

htmlboss:
## Background

As the title states, this is to help avoid any misunderstandings with the visual results, since these ocean features cannot occur at the surface.

* Sonic layer depth
* Deep sound channel axis.

Computationally quickest method that I know of. Just used numpy broadcasting to handle the masking since it's just a fast one-liner.

Nope.

## Screenshot(s)

Sonic Layer Depth:
image

Deep Sound Channel Axis:
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/bin/runserver.sh
+++ b/bin/runserver.sh
@@ -6,6 +6,10 @@
 
 [[ x"$1" == x"" ]] && PORT=5000 || PORT=$1
 
-NUMBER_PROCS=$(awk /processor/'{processor++} END {print processor}' < /proc/cpuinfo)
+if [ ! -e /usr/bin/bc ] ; then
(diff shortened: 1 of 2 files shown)


htmlboss:
Update:
image

htmlboss:
Next steps are to tweak the datasetconfig repository so that the scales are:
* Sonic layer depth: [1.5, 60]
* Deep sound channel axis: [500, 2000]

cc @dwayne-hart

dwayne-hart:
I have updated the datasetconfig.json file to reflect these two new values ranges for the variables requested. I'll open a separate PR for on and off site. So that the changes become persistent.

PR #748, opened 2020-08-10
Updated the runserver.sh script

dwayne-hart:
During my search for ways to improve the user end experience. We should be launching two gunicorn processes per cpu processor found. The update to the runserver.sh script determines the number of processors an OS instance can see in the /proc/cpuinfo file multiplies that value by 2 and launches the number returned from the bc command.

### DIFF
--- a/bin/runserver.sh
+++ b/bin/runserver.sh
@@ -6,6 +6,10 @@
 
 [[ x"$1" == x"" ]] && PORT=5000 || PORT=$1
 
-NUMBER_PROCS=$(awk /processor/'{processor++} END {print processor}' < /proc/cpuinfo)
+if [ ! -e /usr/bin/bc ] ; then
+    NUMBER_PROCS=$(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo)
+else
+    NUMBER_PROCS=$(echo "$(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo) * 2" | bc)
+fi
 
 gunicorn -w $((NUMBER_PROCS)) -t 300 --graceful-timeout 300 --preload -b 0.0.0.0:$((PORT)) --reload "oceannavigator:create_app()" $2


dwayne-hart:
Just a change in the determination of how many gunicorn processes we should launch based on the number of processors seen. From what I have found online we should have a 2:1 ratio for two gunicorn processes to one processor.

dwayne-hart:
One of the articles that i was reading pertained to using [threads](https://pythonspeed.com/articles/gunicorn-in-docker/) and if you were finding that gunicorn was slow to come on-line and you were implementing docker. The authored suggested to instruct the worker to use /dev/shm instead of tmp as some instances do not have a tmp volume available. We fortunately do however, /dev/shm is a faster filesystem than traditional disk.

An [article](https://stackoverflow.com/questions/38425620/gunicorn-workers-and-threads) from 4 years ago provides some strategies for implementing a mixture of work processes and threads. In this particular post, the author uses (2*CPU)+1. 

At present, I have implemented the following code in development and production;

```
NUMBER_PROCS=$(echo "$(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo) * 2" | bc)

gunicorn -w $(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo) --threads=$((NUMBER_PROCS)) \
--worker-class=gthread -t 300 --graceful-timeout 300 --preload -b 0.0.0.0:$((PORT)) --reload  "oceannavigator:create_app()" $2
```

What the above comes out to is if I have an eight (8) processor system. I have eight (8) workers and sixteen (16) per worker. As threads are less resource intensive.

For systems who may not have the `bc` command installed. I have an if statement defined in a development environment...

```
if [ ! -e /usr/bin/bc ] ; then
    NUMBER_PROCS=$(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo)
else
    NUMBER_PROCS=$(echo "$(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo) * 2" | bc)
fi
```

at best you get eight (8) workers and eight (8) threads per worker.

dwayne-hart:
If you want to see what difference this makes. I would encourage all to visit [oceansdata.io/public](https://www.oceansdata.io/public) as it is currently using the following implementation;

```
NUMBER_WORKERS=$(echo "($(awk /^processor/'{processor++} END {print processor}' < /proc/cpuinfo) * 2)+1" | bc)
NUMBER_THREADS=$(echo "${NUMBER_WORKERS} * 2" | bc)

gunicorn -w ${NUMBER_WORKERS} --threads=${NUMBER_THREADS} --worker-class=gthread --worker-tmp-dir /dev/shm -t 300 --graceful-timeout 300 --preload -b 0.0.0.0:$((PORT)) --reload "oceannavigator:create_app()" $2
```

PR #751, opened 2020-08-21
Reformatted the different dataset help files

dwayne-hart:
Used better HTML formatting in the various dataset files.


htmlboss:
could you post a screenshot of one of the rendered files?

PR #756, opened 2020-09-12
Use os path functions for class4 files.

htmlboss:
## Background
Fix class4 functionality breaking whenever root directory locations are changed.

Easiest way I could think of.

Nope.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/class4.py
+++ b/data/class4.py
@@ -22,7 +22,7 @@ def __list_class4_files_slowly():
 
     for f in glob.iglob(f"{class4_path}/**/*.nc", recursive=True):
         if f.endswith('profile.nc') and ('GIOPS' in f):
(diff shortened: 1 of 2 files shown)


data/class4.py
```diff
@@ -22,7 +22,7 @@ def __list_class4_files_slowly():
 
     for f in glob.iglob(f"{class4_path}/**/*.nc", recursive=True):
         if f.endswith('profile.nc') and ('GIOPS' in f):
-            files.append(f[18:-3])
+            files.append(os.path.splitext(os.path.basename(f))[0])
```
htmlboss, line 25: This method should have been used in the first place.

PR #765, opened 2020-10-08
Delete unused utils/log.py module

douglatornell:
## Background
The `utils/log.py` module is unused code. It is not imported anywhere in the codebase, nor are there any calls to the functions it contains anywhere. 

Deleting unused code is a Good Thing™.

Reading the code in `utils/log.py` makes me think that it pre-dates the Flask API in Navigator. Logging is now set up by Flask and the `logger` object is now accessed as `flask.current_app.logger`. In contrast, `utils/log.py` uses the Python `logging` module directly to set up a `logger`, but there is now no way for that `logger` to be accessible.

I have run my dev instance for several hours with the `utils/log.py` module deleted and have not encountered any exceptions due to its absence.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


htmlboss (approved):
💥

PR #766, opened 2020-10-14
Fix class4 model list

htmlboss:
the paths changed on hank so it broke

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


dwayne-hart:
This evening I checked out the updated main branch. Unfortunately, after performing an indexing of the CLASS4 data files only the CONCEPTS model is available for comparison with the real world observations.

```
commit 6d8ef18959bafaaa6312dcd5df068b78ff0b043f (HEAD -> master, origin/master, origin/HEAD)
Author: Nabil <nmiri127@gmail.com>
Date:   Mon Oct 19 09:26:56 2020 -0230

    Fix class4 model list (#766)
    
    the paths changed on hank so it broke

```
image

dwayne-hart:
I should have also mentioned that the new directory structure of the CLASS4 files is the following;

`/data/class4/{YYYY}/{YYYYMMDD}/class4_{YYYYMMDD}_{SLA,SST,aice,profile}.nc`

For the April 8, 2020 plot (shown above) the following models file are available;

image

Even with creating symbolic links on the files in `/data/class4/{YYYY}` to point into the below subdirectories did not provide the desired outcome.

VanessaSutton-Pande:
It sounds like additional work is required from the code base for this fix (other global datasets). As long as we have the data feeds stabilized and new data is displayed in production site then let's park this one for now. If we get specific feedback from end users then we can then re-examine.

PR #767, opened 2020-10-19
Changing the size and color used when plotting the bathymetry dataset.

dwayne-hart:
## Background

A number of end users and stakeholders have requested that the size of the font being used when the bathymetry data is viewed be increased and the color to be a more prominent.

The font size change was easier than selecting the color. In the end I removed the array values and used a direct call to a color scheme.

## Screenshot(s)

image

## Checks
- [* ] I ran unit tests.
- [*] I've tested the relevant changes from a user POV.

```sh
bash run_python_tests.sh
```
Here are the results of running the unit tests;

```
======================================================================
ERROR: test_deepsoundchannel (tests.test_data_functions.TestDataFunctions)
----------------------------------------------------------------------
  File "/home/buildadm/Ocean-Data-Map-Project/tests/test_data_functions.py", line 47, in test_deepsoundchannel
    data=[0], dims=['depth']), [45], [0.5], [32]), 0)
  File "/home/buildadm/Ocean-Data-Map-Project/data/calculated_parser/functions.py", line 334, in deepsoundchannel
    data[data < 500] = np.nan
TypeError: 'numpy.int64' object does not support item assignment

----------------------------------------------------------------------
Ran 183 tests in 1586.729s
FAILED (errors=1, skipped=20)
```

### DIFF
--- a/plotting/map.py
+++ b/plotting/map.py
@@ -667,12 +667,9 @@ class MapPlotter(Plotter):
                 self.longitude, self.latitude, self.bathymetry, latlon=True,
                 linewidths=0.5,
                 norm=LogNorm(vmin=1, vmax=6000),
-                cmap=mcolors.LinearSegmentedColormap.from_list(
-                    'transparent_gray',
-                    [(0, 0, 0, 0.5), (0, 0, 0, 0.1)]
-                ),
+                cmap='Greys',
                 levels=[100, 200, 500, 1000, 2000, 3000, 4000, 5000, 6000])
-            plt.clabel(cs, fontsize='xx-small', fmt='%1.0fm')
+            plt.clabel(cs, fontsize='x-large', fmt='%1.0fm')
 
         if self.area and self.show_area:
             for a in self.area:
--- a/plotting/map.py
+++ b/plotting/map.py
@@ -3,7 +3,6 @@ import os
 import tempfile
 from textwrap import wrap
 
-import matplotlib.colors as mcolors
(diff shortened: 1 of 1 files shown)


dwayne-hart:
After removing the unused `import of import matplotlib.colors as mcolors`

image

htmlboss (commented):
Size looks good. Colour is improved but I'm wondering should we try black?

dwayne-hart:
The color appears to scale depending on the parent colorbar. The second pic
shows a black font in a red field.

Best,
Dwayne

nsoontie:
Thanks for tackling this Dwayne! It looks good.

Quick question for @htmlboss : Is the failed test something to be concerned about?

As for colours, the contour colour changes with depth. Deeper = blacker. To choose solid black color, you could use colors='k' instead of cmap='Greys' in the self.basemap.contour(..) call.

https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/c8b8e8bac418d6815ead4ae352ed9307a39f6c82/plotting/map.py#L665-L670

htmlboss:
Nothing to worry about. Looks like codebeat fell on its face when talking to github: https://codebeat.co/projects/github-com-dfo-ocean-navigator-ocean-data-map-project-master/pull_requests/399369

dwayne-hart:
Its been like that since last night.

D

htmlboss (approved):

Issue #769, opened 2020-10-22
Issue with exporting to NetCDF-3 NC


dwayne-hart:
**Describe the bug**
When selecting an area of interest the plot renders and when we wish to export any variable(s) to a NetCDF-3 NC format. We receive an Internal Error.

**To Reproduce**
Request any area and variable(s) combination to export to NetCDF-3 NC. The error will be experienced immediately.

**Expected behavior**
Generation of a NetCDF-3 NC formatted file for post analysis.

**Screenshots**
![image](https://user-images.githubusercontent.com/44408505/96874374-a0a46680-1450-11eb-8e3b-400d853ac70e.png)

![image](https://user-images.githubusercontent.com/44408505/96874465-b580fa00-1450-11eb-89ec-ffe172e48964.png)

**Desktop (please complete the following information):**
 - OS: MacOS, Windows 10 Enterprise 
 - Browser Chrome
 - Version Version 85.0.4183.121 (Official Build) (64-bit)

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

dwayne-hart:
Update, the following information was taken from a test container which has the same issue as production when we try to export variables from the GIOPS 10 day Forecast 3D - LatLon

```
[2020-10-22 11:16:35 -0230] [4661] [ERROR] Error handling request /api/v1.0/subset/?&output_format=NETCDF3_NC&dataset_name=giops_day&variables=votemper,vozocrtx,vomecrty,magwatervel&min_range=41.82992111131577,-56.591606140136726&max_range=48.345069092460676,-47.261466979980476&time=2235297600,2235297600&user_grid=0&should_zip=0
Traceback (most recent call last):
  File "/home/buildadm/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/gunicorn/workers/gthread.py", line 279, in handle
    keepalive = self.handle_request(req, conn)
  File "/home/buildadm/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/gunicorn/workers/gthread.py", line 328, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/home/buildadm/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
[...]

PR #774, opened 2020-10-30
Fix ValueError due to 1D lons/lats in subset API call

douglatornell:
## Background
Issue #769 revealed 2 ways in which a `subset` API call can result in a `ValueError`. This PR addresses one of them. A subsequent PR will address the other. 

In the case addressed here, @nsoontie identified in https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/769#issuecomment-715537549 that the `ValueError` is due to the lons and lats variables being 1D arrays but `pyresample.geometry.SwathDefinition()` requires 2D arrays. The solution that @nsoontie came up with is to use `numpy.meshgrid()` to generate 2D lons and lats arrays.

It's a very simple, elegant 2-line solution to the problem.

It probably took me longer to cook up the unit test that demonstrates that the problem is resolved than it took @nsoontie to come up with the solution :smile: 

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -595,7 +595,7 @@ class NetCDFData(Data):
                 return pyresample.kd_tree.resample_nearest(input_def, data,
                                                            output_def, radius_of_influence=float(self.radius), nprocs=8)
 
-        raise ValueError(f"Unknown interpolation method {self.interp}.")                                               
+        raise ValueError(f"Unknown interpolation method {self.interp}.")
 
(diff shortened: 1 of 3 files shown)


nsoontie (approved):
Looks good Doug! And thanks for adding the unit test :)

htmlboss (approved):

PR #775, opened 2020-11-04
Change showmap default to false for Points & Class4 UI overlays

douglatornell:
## Background
In the hunt for ways to improve the performance of the profile plots in the Points UI overlay I found that the map generated when the Show Location checkbox is checked takes a significant percentage of the total rendering time. The changes in this PR make the default for Show Location in the Points and Class3 overlays unchecked.

I also considered changing the default checked setting of the Show Map checkbox in the Lines and Observation Tracks overlays, but decided that the map tiles in those overlays provide sufficiently useful information that they should remain. Please see the notes on the screenshots below. **I would like confirmation of that decision from people who are closer to the user community than I am.**

The backend code that runs when `showmap` is `true` takes 2 to 3 seconds, which is 30%-50% of the total time to generate the content of the Points and Class4 overlays. In those overlays the graphic produced when `showmap` is `true` is a fairly uninformative single point on a map tile. Speeding up the generation of the overlays content at the expense of not showing those map tiles seems like a good trade-off to improve UX, and @VanessaSutton-Pande's recent discussion with users of that trade-off got their agreement to go ahead with it.

@dwayne-hart Please note that deployment of this change requires a rebuild of the frontend components in the production container.

## Screenshot(s)
Points overlay with this PR's change in place:
ProfileShowLocationUnchecked

Class4 overlay with this PR's change in palce:
Class4ShowLocationUnchecked

Lines overly Show Map Location provides useful information about the direction of the transect, so it remains checked by default:
LineShowMapLocationChecked

Observations Track overlay Show Map provides useful information about the shape of the the track, so it remains checked by default:
ObservarionsShowMapChecked

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Class4Window.jsx
+++ b/oceannavigator/frontend/src/components/Class4Window.jsx
@@ -17,7 +17,7 @@ export default class Class4Window extends React.Component {
 
     this.state = {
       forecast: "best",
-      showmap: true,
+      showmap: false,
       climatology: false,
       error: "none",
       size: "10x7",
--- a/oceannavigator/frontend/src/components/Class4Window.jsx
+++ b/oceannavigator/frontend/src/components/Class4Window.jsx
@@ -43,11 +43,11 @@ export default class Class4Window extends React.Component {
 
   onLocalUpdate(key, value) {
     if (this._mounted) {
-     
+
       let newState = {};
       if (typeof(key) === "string") {
         newState[key] = value;
-      } 
+      }
       else {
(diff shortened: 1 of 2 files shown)


dwayne-hart:
Awesome work @douglatornell . Just for everyone's benefit the production containers are replaced entirely with new instances with everything built and tested before an IP flip. The older containers are removed after 7 to 14 days if we need to revert.

htmlboss (approved):

nsoontie (approved):

PR #776, opened 2020-11-04
Add Sentry integration for backend via Flask app

douglatornell:
## Background
This is the setup required to facilitate capture and analysis of Python exceptions by Sentry. It is enabled in a container by providing an environment variable named `SENTRY_DSN` whose value is a URL generated by Sentry for the "project" on their service that is to receive the data. If the envvar is absent or empty, the Sentry integration does nothing. So, it is safe to have this code in the codebase that is used for dev or any deployment.

It's the documented way to do it: https://docs.sentry.io/platforms/python/guides/flask/

TODOs before this can go live:
- [x] Add the `sentry-sdk` and `blinker`[1] packages to the conda env config. `conda install -c conda-forge sentry-sdk blinker` did the job in my dev container on `dory`. It installed `sentry=0.19.1` and `blinker=1.4` and updated `certifi` to `2020.06.20`. @dwayne-hart do you mind doing the the conda environment and spec-file changes for this? I think you have a process for those kind of changes that I would likely mess up if I were to do them. Your choice whether you do a separate PR, or commit the changes in this branch.

  [1] `blinker` (https://pythonhosted.org/blinker/) is used by the Flask signal component that `sentry-sdk` hooks into.

- [x] Chose a name for the Sentry project that we want to capture exception data from https://navigator.oceansdata.ca/public/ in. `navigator-prod` maybe? Opinions @dwayne-hart and @htmlboss ?  _Decided on `public-prod`_.

- [x] Once we have a Sentry project name I will set it up and that will generate the DSN URL that will enable data capture. _See https://sentry.io/settings/dfo-ocean-navigator/projects/public-prod/keys/_

- [x] Figure out how to ensure that the `SENTRY_DSN` envvar is always set for the https://navigator.oceansdata.ca/public/ container while keeping the DSN URL secret (i.e. not committed). Ideas @dwayne-hart ?

- [ ] Deploy

- [x] Profit!? :grin: 

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/__init__.py
+++ b/oceannavigator/__init__.py
@@ -4,6 +4,8 @@ from sys import argv
 from flask import Flask, request, send_file
 from flask_compress import Compress
 from flask_babel import Babel
+import sentry_sdk
+from sentry_sdk.integrations.flask import FlaskIntegration
 
 # Although DatasetConfig is not used in this module, this import is absolutely necessary
 # because it is how the rest of the app gets access to DatasetConfig
@@ -17,6 +19,8 @@ def config_blueprints(app):
     app.register_blueprint(bp_v1_0)
 
 def create_app(testing = False):
+    # Sentry DSN URL will be read from SENTRY_DSN environment variable
+    sentry_sdk.init(integrations=[FlaskIntegration()], traces_sample_rate=1.0)
     app = Flask(__name__, static_url_path='', static_folder='frontend')
     app.add_url_rule('/', 'root', lambda: app.send_static_file('index.html'))
     app.config.from_pyfile('oceannavigator.cfg', silent=False)
--- a/routes/api_v1_0.py
+++ b/routes/api_v1_0.py
@@ -72,6 +72,14 @@ def info_v1_0():
         "This is the Ocean Navigator API - Additional Parameters are required to complete a request, help can be found at ...")
 
 
+@bp_v1_0.route('/api/test-sentry')
+def test_sentry():
+    # Hit this endpoint to confirm that exception and transaction logging to Sentry are
+    # operating correctly; a transaction should appear in the appropriate project at:
+    # https://sentry.io/organizations/dfo-ocean-navigator/performance/
+    raise APIError("This is the Ocean Navigator API Sentry integration test endpoint.")
+
+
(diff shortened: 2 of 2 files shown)


dwayne-hart:
@douglatornell, if we have three or more LXD/LXC containers running in production. Should we have all three or more sending messages to the same project? Is there a way to create a subproject within production for each node?

I would ask that three projects be created and called them dev, prod, bug.

douglatornell:
> @douglatornell, if we have three or more LXD/LXC containers running in production. Should we have all three or more sending messages to the same project? 

@dwayne-hart, my initial thought is that if 3 or more containers are running the same version of the code, that would be fine. But if they are running different codebases (dev, prod, bug) things could be confusing.

> Is there a way to create a subproject within production for each node?

There is a Sentry feature called Environments that I haven't used and need to read about that might be like subproject - not sure yet.

> 
> I would ask that three projects be created and called them dev, prod, bug.

I'm thinking we do `prod` for the https://navigator.oceansdata.ca/public/ container to start with so that we can some immediate info about the exceptions that our users are triggering. We can expand to more projects/environments as we get more experience with Sentry and figure out how best to use it to improve Navigator. Does that make sense, or am I too much of an incrementalist?

So maybe Sentry project name `prod` or `public-prod`? I'm thinking about other "released production deployments" like GPSC maybe that would need an `adjective-prod` name. I really don't know enough about the deployment side though, so maybe I'm typing nonsense.

It's late there - don't feel that you have to continue this conversation tonight!

dwayne-hart:
@douglatornell , I agree with your strategy wrt the slow roll on the prod or public-prod. Having the ability to create separate environments. Maybe we would be able to have an environment for each deployment we maintain. If we wish for others to contribute we need to reach out to the various stakeholders who maybe running their own instance if they wish to opt in. We can definitely discuss this further whenever you wish.

dwayne-hart:
A new navigator spec file has been created #781 which has the supporting packages for sentry.io. Also, a new Ubuntu 20.04 LTS image was created to have the necessary environment variable define.

dwayne-hart (approved):
This work will help with understanding where things break down via sentry.io.

PR #779, opened 2020-11-09
Zarr data access enabled

CubeCode93:
## Background
To enable ocean navigator to access data from Zarr data structure 

As a part of optimization, we observed that loading data from Zarr takes less time than netCDF

Need to install Zarr package 
_> conda install -c conda-forge zarr_

By default API calls will work with netCDF, if we want to use Zarr than change [giops_day => giops_day_zarr]  in API call.
Eg: 
http://navigator.oceansdata.ca/api/v1.0/tiles/gaussian/25/10/EPSG:3857/**giops_day_zarr**/votemper/2236852800/0/-5,30/4/5/5.png

Define the zarr file path at datasetconfig.json file under,
"giops_day_zarr": {
     "url_zarr": "/data/zarrfile.zarr"
},

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


nsoontie:
Hi @CubeCode93 
We're looking at discussing your pull request during the technical working group call in the coming weeks. In the meantime, a few things for you to think about
1.  Do all of the unit tests still pass?
You can check with `bash run_python_tests.sh`
2. Have you contributed new unit tests for your changes? Are there any new tests for the ZarrData class?
3. Does this pull request enable use of both ZarrData and NetCDFData? Or is it one or the other?

CubeCode93:
> Hi @CubeCode93
> We're looking at discussing your pull request during the technical working group call in the coming weeks. In the meantime, a few things for you to think about
> 
> 1. Do all of the unit tests still pass?
>    You can check with `bash run_python_tests.sh`
I ran the unit test, it's doesn't show any error related to my changes. I am attaching the unit test output below.
unit_test_output

> 2. Have you contributed new unit tests for your changes? Are there any new tests for the ZarrData class?
No. I haven't, basically ZarrData class is a clone of NetCDF class with some changes to load Zarr data.

> 3. Does this pull request enable use of both ZarrData and NetCDFData? Or is it one or the other?
It supports both, by changing the dataset parameter in API call a user can decide whether to go with netCDF or Zarr
Eg: In API call, if we use "giops_day" it loads from netCDF and if we call "giops_day_zarr" it loads from Zarr
netCDF:
http://navigator.oceansdata.ca/api/v1.0/tiles/gaussian/25/10/EPSG:3857/**giops_day**/votemper/2236852800/0/-5,30/4/5/5.png
Zarr:
http://navigator.oceansdata.ca/api/v1.0/tiles/gaussian/25/10/EPSG:3857/**giops_day_zarr**/votemper/2236852800/0/-5,30/4/5/5.png

Note: As of now Zarr feature works for GIOPS_DAY dataset only

jmunroe:
I agree we need discuss these changes in more detail before merging.  I'd like to distinguish between adding "zarr" as a fileformat versus any changes to the dataset config.  In particular, this PR needs to be completely independent from any changes to datasetconfig.json

The most significant change is you seem to subclassing CalculatedData from ZarrData instead of NetcdfData.  This sounds like a really major change.  

Second, how much code duplication is there between NetcdfData and ZarrData ?  Your new file "zarr_data.py" is really big. Do you really need to add all that new code?

nsoontie:
I agree with @jmunroe. While we may want to enable the use of zarr data but we don't want to replace all of our datasets with zarr data at the same time. 

@dwayne-hart Do you have any ideas about the missing class4_files.pickle error in @CubeCode93's attempt to run the tests?

dwayne-hart:
To generate a pickle file containing the location of all the CONCEPTS
CLASS4 files. You will need to ensure the Miniconda toolchain is part of
your current sessions PATH environment variable and then run the
`./Ocean-Data-Map-Project/bin/launch-pickle.sh` script. For the instances
we have in production, I used the following command as they are set up
differently than what   @CubeCode93 <https://github.com/CubeCode93> may be
using in their deployment.

```
source tools/conf/ocean-navigator-env.sh ;
./Ocean-Data-Map-Project/bin/launch-pickle.sh
```

data/netcdf_data.py
```diff
@@ -53,6 +54,7 @@ def __enter__(self):
         if not self.meta_only:
             # Don't decode times since we do it anyways.
             decode_times = False
+            url = self.url if not isinstance(self.url, list) else self.url[0]
```
jmunroe, line 57: Is this zarr specific? Probably should be incorporated into line 71 directly.  Careful -- what would happen if "url" was a list?

data/netcdf_data.py
```diff
@@ -635,14 +642,19 @@ def dimensions(self) -> List[str]:
                     dimension_list = db.get_all_dimensions()
             except sqlite3.OperationalError:
                 pass
-            return dimension_list
-
+        
+        elif url.endswith(".zarr"):
```
jmunroe, line 646: Again, what would happen here if `url` was indeed a list?

data/netcdf_data.py
```diff
@@ -718,22 +730,39 @@ def variables(self) -> VariableList:
...
+            ds_zarr = xarray.open_zarr(url)
+            var_list =[]
+            for var in list(ds_zarr.data_vars):
+
+                name = var
+                units = ds_zarr.variables[var].attrs['units'] if ds_zarr.variables[var].attrs['units'] else None
+                long_name = ds_zarr.variables[var].attrs['long_name'] if ds_zarr.variables[var].attrs['long_name'] else name
+                valid_min = ds_zarr.variables[var].attrs['valid_min'] if ds_zarr.variables[var].attrs['valid_min'] != 1.17549e-38 else None
+                valid_max = ds_zarr.variables[var].attrs['valid_max'] if ds_zarr.variables[var].attrs['valid_max'] != 3.40282e+38 else None
```
jmunroe, line 743: The comparison against "exact" floating point values doesn't make sense.  if attrs['valid_min'] does not exist, then valid_min should be None. Otherwise, leave valid_min as is (and likewise for valid_max)

data/netcdf_data.py
```diff
@@ -718,22 +730,39 @@ def variables(self) -> VariableList:
...
+
+                var_list.append(Variable(name, long_name, units, list(itertools.chain(*[ds_zarr[name].dims])), valid_min, valid_max))
```
jmunroe, line 745: is `list(iterools.chain(*[ds_zarr[name].dims))` really needed? Isn't `list(ds_zarr[name.dims])` sufficient?

tests/test_netcdf_data.py
```diff
@@ -48,6 +49,36 @@ def test_enter_nc_files_list(self):
...
+                                "long_name" : "Sea water potential temperature",
+                                "valid_min" : 173.0,
+                                "valid_max" : 373.0,
+                                }
+                            ),
+                    }
+        )
+    
+        if not (os.path.exists("tests/testdata/giops_test.zarr")):
+            ds.to_zarr("tests/testdata/giops_test.zarr")
```
jmunroe, line 74: Thanks for creating a test data on the fly.  But is the the recommended way of creating test data in the testing framework? Do we need to clean up this file after the test? What about the other test_zarr_* tests --- do we not need to create the test file first for those as well?

There should be someway of doing up a "setup" and "teardown" within the testing framework.

data/netcdf_data.py
CubeCode93: Yes, It's specific to Zarr.
Now it's added into line 71 (New line:69).
If the URL is given as a list for other file formats, IF condition will handle it, hope Zarr file path won't be a list

data/netcdf_data.py
CubeCode93: If it's a list, it will be handled by line (new): 635

data/netcdf_data.py
CubeCode93: Done.

data/netcdf_data.py
CubeCode93: Yes, it is.
Modified the code accordingly in the new commit.

tests/test_netcdf_data.py
CubeCode93: I don't see any conflicts with other test methods by creating test data in the testing framework.
The latest commit won't require us to clean the data file, it creates and delete the testing data after the tests.

jmunroe (approved):
Great. I think this now addresses any concerns I had and makes it much clearer what the PR is doing.  nice to see the setUp and tearDown routines.

jmunroe:
I think it is now ok to release the draft status on this PR.  @htmlboss Did you want to look at this before the PR gets merged?

data/netcdf_data.py
```diff
@@ -65,13 +65,18 @@ def __enter__(self):
                     # xarray won't open FVCOM files due to dimension/coordinate/variable label
                     # duplication issue, so fall back to using netCDF4.Dataset()
                     self.dataset = netCDF4.MFDataset(self._nc_files)
+
+            elif (self.url.endswith(".zarr") if not isinstance(self.url, list) else False):
+                ds_zarr = xarray.open_zarr(self.url, decode_times=decode_times)
+                self.dataset = ds_zarr
+
```
douglatornell, line 72: I'm sorry to have to say this, @CubeCode93 , but the changes I made in PR #789 that got merged this week are in this area of the code, so you will have to merge with HEAD again in you branch. It might be a bit messy for you, I'm afraid. Sorry.

douglatornell (changes requested):
Looks good to me!
+1 on using the `setUp()` and `tearDown()` methods to manage test datasets on-the-fly.
Sorry about making `netcdf_data.py` a moving target for you (noted below). I will keep my hands off it until after this PR is merged.

data/netcdf_data.py
CubeCode93: Thanks for the review and comments @douglatornell 
That's totally fine and I have updated my branch, hope it's fine now!

douglatornell (approved):

dwayne-hart:
@CubeCode93 please update your branch and I will gladly merge in your work.

CubeCode93:
> @CubeCode93 please update your branch and I will gladly merge in your work.

Updated.

PR #780, opened 2020-11-10
Fix ValueError due to "dimensions do not exist" in subset API call

douglatornell:
## Background
Issue #769 revealed 2 ways in which a `subset` API call can result in a `ValueError`. PR #774 addressed one of them. This PR addresses the other. 

In the case addressed here, @nsoontie identified in https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/769#issuecomment-715598904 that the `ValueError` is due to the code trying to use longitude/latitude variables names in `Dataset.isel()` when it should be using x/y dimensions.

The solution Implemented in this PR was proposed by @nsoontie. It adds properties to the `NetCDFData` class to facilitate finding y and x dimension `DataArrays` in `Datasets` with a variety of dimensions via set intersections. 

1. Adding properties that provide a set of possible x/y dimension names gives us a single place to add dimension names from future data sources.
2. Intersecting the set of possible dimension names with the `Dataset.dimensions` cast to a set is fast and elegant.

I believe that issue #769 can be closed when this PR is merged.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -4,7 +4,7 @@ import sqlite3
 import uuid
 import warnings
 import zipfile
-from typing import List, Dict, Union
+from typing import List, Dict, Union, Set, Tuple
 
 import dateutil.parser
 import geopy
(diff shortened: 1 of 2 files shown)


nsoontie (approved):
Looks great Doug! Thanks for adding the unit tests and the properties listing possible y dimensions and x dimensions. 

I'm curious about your thoughts too @htmlboss.

htmlboss (approved):

PR #781, opened 2020-11-13
Update for the navigator python virtual environment

dwayne-hart:
## Background

In order to support the work @douglatornell has undertaking in bringing sentry.io api error logging and metrics. Please refer to #776. The Miniconda toolchain has had the sentry-sdk and blinker packages installed. A new conda spec file for the navigator virtual environment was also created in order to ensure that any Ocean Navigator deployment will be able to receive not only the package artifacts but the spec file to be able to create the Miniconda toolchain from scratch.


htmlboss:
How do the unit tests fare with the spec file changes?

PR #788, opened 2020-11-24
Add possible Sentry environment name from envvar

douglatornell:
## Background
This change adds tagging of exception events captured by Sentry with and environment name that is read from the `SENTRY_ENV` environment variable. If that envvar is unset, the exceptions are not tagged, but they are still captured by Sentry.

Using an environment variable and the fact at the value returned by `os.getenv()` defaults to `None` gives us the flexibility to deploy Navigator instances with `SENTRY_ENV` not set with no negative consequences. When `SENTRY_ENV` is set, exception events captured by Sentry will be tagged with the environment name, enabling easy filtering of events from different environments; e.g. NAFC and `dory` production containers.

I propose that we use `NAFC` and `dory-prod` as environment names for the two deployments where we presently have Sentry activated.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/__init__.py
+++ b/oceannavigator/__init__.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from sys import argv
 
 from flask import Flask, request, send_file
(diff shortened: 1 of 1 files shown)


dwayne-hart:
I would go even further and use DAL or DAL-Prod and DAL-Dev instead of dory-prod. IMHO.

douglatornell:
`DAL-Prod` works for me.

dwayne-hart:
I'll put the variable in play later today for NAFC.

dwayne-hart (approved):
This is a clean and simple change to leverage an environment variable for sentry to parse the incoming issues.

Issue #690, opened 2020-03-19
Figure out optimal chunk size with open_mfdataset


htmlboss:
#689 makes a first cut at figuring out an optimal chunk size of 200. But the question stands, is there a more optimal number?

PR #789, opened 2020-11-25
Only use open_mfdataset() when there is >1 dataset

douglatornell:
## Background
There is a measurable performance cost to using `xarray.open_mfdataset()` with
chunking instead of `xarray.open_dataset()` without chunking when we are only
opening a single dataset.

Timing statistics for `routes.api_v1_0.tile_v1_0()` calls for single dataset variables initial page load with cold caches:

`xarray.open_mfdataset(..., chunks=200)`:
Statistic | Value
-- | --
count | 61
min | 0.817 s
mean | 2.496 s
median | 1.854 s
std dev | 1.679 s
max | 7.855 s

`xarray.open_dataset()` with no chunking:
Statistic | Value
-- | --
count | 61
min | 0.578 s
mean | 1.515 s
median | 1.451 s
std dev | 0.411 s
max | 2.491 s

I don't believe that the uniform chunking factor of 200 on all dimensions that is presently hard-coded is optimal when we are loading multiple datasets (either to retrieve multiple variables, or the retrieve long time series). Chunking for data access should be related to the chunking scheme that is used for the data storage, the memory available for `dask` to operate in, and the dimensionality of the data use. All of those issues need to be considered in addressing issue #690.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -56,11 +56,17 @@ class NetCDFData(Data):
 
             if self._nc_files:
                 try:
-                    self.dataset = xarray.open_mfdataset(
-                        self._nc_files,
-                        decode_times=decode_times,
-                        chunks=200,
-                    )
+                    if len(self._nc_files) > 1:
+                        self.dataset = xarray.open_mfdataset(
+                            self._nc_files,
+                            decode_times=decode_times,
+                            chunks=200,
+                        )
+                    else:
+                        self.dataset = xarray.open_dataset(
+                            self._nc_files[0],
+                            decode_times=decode_times,
+                        )
                 except xarray.core.variable.MissingDimensionsError:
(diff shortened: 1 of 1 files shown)


htmlboss (approved):
The difference in std dev makes me wonder what is dask doing under the hood that's causing such a variable delay. Anyways, great catch!

htmlboss:
Codefactor is drunk, you can ignore it if you wish lol

douglatornell:
> Codefactor is drunk, you can ignore it if you wish lol

Agreed. None of those issues are in the piece of code this PR touches. The only one that makes slight sense is the complaint about a bare `except`, and that's not a blocker.

PR #791, opened 2020-12-01
Drop unused `pint.UnitRegistry` in `Mercator.depths` & `Nemo.depths` properties

douglatornell:
## Background
Instantiation of `pint.UnitRegistry` appears to take ~200ms every time the `depths` property is accessed.
However, code inspection and UI testing show that none of the features of `UnitRegistry` are being used on `depths`;
i.e. there are no accesses of `depths.magnitude` or `depths.units`, nor `pint` unit conversions on `depths`.

The `depths` property is accessed by several API endpoints, notably `tile_v1_0`, so eliminating that ~200ms can produce noticeably snappier performance on pages for which the caches are cold.

Timing statistics for `Mercator.depths` with `pint.UnitRegistry()` for single dataset variables initial page load with cold caches:
Statistic | Value
-- | --
count | 61
min | 0.142
mean | 0.253
median | 0.216
std dev | 0.110
max | 0.578

and without `pint.UnitRegistry()`:
Statistic | Value
-- | --
count | 61
min | 0.00046
mean | 0.00185
median | 0.00062
std dev | 0.00376
max | 0.01604

Timing statistics for `routes.api_v1_0.tile_v1_0()` calls with `pint.UnitRegistry()` for single dataset variables initial page load with cold caches:
Statistic | Value
-- | --
count | 61
min | 0.728
mean | 1.542
median | 1.476
std dev | 0.529
max | 3.921

and without `pint.UnitRegistry()`:
Statistic | Value
-- | --
count | 61
min | 0.750
mean | 1.368
median | 1.195
std dev | 0.491
max | 2.457

The above stats were collected in a context that also includes PR #789.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/mercator.py
+++ b/data/mercator.py
@@ -48,14 +48,8 @@ class Mercator(Model):
                     break
             self.__depths = var.values if var is not None else np.array([0])
 
-            if var is not None:
-                ureg = UnitRegistry()
-                unit = ureg.parse_units(var.attrs['units'].lower())
-                self.__depths = ureg.Quantity(var[:].values, unit).to(ureg.meters).magnitude
-            else:
-                self.__depths = np.array([0])
-
-            self.__depths.flags.writeable = False
+            # Make immutable
+            self.__depths.setflags(write=False)
 
         return self.__depths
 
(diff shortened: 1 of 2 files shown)


dwayne-hart:
Removing the unused calculations results is a good approach.

dwayne-hart (approved):
Approving this week.

htmlboss (approved):

PR #793, opened 2020-12-04
Delete unused UnitRegistry imports

douglatornell:
## Background
This should have been done in PR #791, but I missed it and didn't see that LGTM had flagged it before the PR got merged :frowning_face: 

Unused imports are distracting noise in the context of code comprehension.

No.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/mercator.py
+++ b/data/mercator.py
@@ -2,7 +2,6 @@ from typing import Union
 
 import numpy as np
 import pyresample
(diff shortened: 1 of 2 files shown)


dwayne-hart (approved):
Will update staging shortly.

PR #794, opened 2020-12-06
Create codeql-analysis.yml

htmlboss:
Automated code vulnerability analysis for JS and Python.

File was auto-generated by github.

Nope

## Screenshot(s)
n/a

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


dwayne-hart (approved):
Very interesting. This should provide benefits to both the front and middle layers.

Issue #796, opened 2020-12-08
/api/v1.0/class4/models/{class4_id} always returns empty list


douglatornell:
**Describe the bug**
The `/api/v1.0/class4/models/{class4_id}` endpoint always returns an empty list, resulting in no "Additional Models" selector section in the "Class 4 Settings" panel of the overlay (please see screenshot below).

**To Reproduce**
Steps to reproduce the behaviour:
* API endpoint option:
  * Send a GET request to `api/v1.0/class4/models/class4_20191217_GIOPS_CONCEPTS_3.0_profile_253/` 
  * The result return is a JSON object containing an empty list
* UI option:
  * Choose 2019-12-17 in the Class4 date selector
  * Click on one of the Class 4 location points
  * There is no "Additional Models" selector section in the "Class 4 Settings" panel of the overlay (please see screenshot below)

**Expected behaviour**
* API endpoint option:
  * The result is a JSON object like:
    ```json
    [{"id":"class4_20191217_BLK_omaps_3.2_profile","value":"BLK"},
     {"id":"class4_20191217_FOAM_orca025_14.1_profile","value":"FOAM"},
     {"id":"class4_20191217_PSY3V3R3_orca025_profile","value":"PSY3V3R3"},
     {"id":"class4_20191217_PSY4V3R1_orca12_profile","value":"PSY4V3R1"}]
    ```
* UI option:
  * There should be an "Additional Models" selector section in the "Class 4 Settings" panel of the overlay

**Screenshots**
Missing "Additional Models" selector:
![image](https://user-images.githubusercontent.com/76797/101548361-90910980-3960-11eb-91a1-7a37f2d74cc7.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 20.04
 - Browser [e.g. chrome, safari] firefox
 - Version [e.g. 22] 83.0
No reason to suspect that this does not occur on all OSs and browser versions.

**Smartphone (please complete the following information):**
I didn't check, but there is no reason to suspect that the issue does not arise on mobile too.

**Additional context**
N/A

PR #797, opened 2020-12-08
Fix for issue #796 re: class4/models/{class4_id} API endpoint

douglatornell:
## Background
re: issue #796 The `/api/v1.0/class4/models/{class4_id}` endpoint always returns an empty list, resulting in no "Additional Models" selector section in the "Class 4 Settings" panel of the overlay.

This fix makes the code compatible with the `/data/class4/yyyy/yyyymmdd/` files storage scheme that is in use.

I also took the opportunity to modernize the code to use `pathlib`.

## Screenshot(s)
Broken:
image

Fixed:
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/class4.py
+++ b/data/class4.py
@@ -4,6 +4,7 @@ import glob
 import os
 import pickle as pickle
 import time
+from pathlib import Path
 from typing import List
 
 import cftime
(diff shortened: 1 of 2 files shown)


dwayne-hart (approved):
This looks really good @douglatornell .

PR #799, opened 2020-12-09
Fix for issue #798 re: FileNotFoundError in class4 forecasts API request

douglatornell:
## Background
re: GitHub issue #798 and Sentry issue # PUBLIC-PROD-4

The class4 id in the API request ends with an index number that `data.class4.list_class4.forecasts()` is not expecting. Fixed that parsing by stripping of the index number. That revealed an [AttributeError issue](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/798#issuecomment-742164788) a few lines later. It was resolved by adding `decode_times=False` to xarray.open_dataset(), consistent with how we open datasets in `netcdf_data.NetCDFData.__enter__()` because the netCDF times are decoded in our code.

Only way to make the code work.

Added unit test for part of `data.list_class4_models()` functionality.

The that generates the list of forecast dates for the frontend to use can't possibly add anything to the `result` list because the values in the `modeljuld` variable in the class 4 dataset are all always the observation date. The `leadtime` variable needs to be incorporated into the calculation. But that is all beyond the scope of the issue that this PR is addressing.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/class4.py
+++ b/data/class4.py
@@ -215,10 +215,10 @@ def list_class4_forecasts(class4_id: str) -> List[dict]:
     """
     dataset_url = current_app.config["CLASS4_FNAME_PATTERN"] % (
         class4_id[7:11], class4_id.rsplit('_', maxsplit=1)[0])
-    with xr.open_dataset(dataset_url) as ds:
-        var = ds['modeljuld']
-        forecast_date = [d.strftime("%d %B %Y") for d in
-                         cftime.utime(var.units).num2date(var[:])]
+    with xr.open_dataset(dataset_url, decode_times=False) as ds:
+        forecast_date = [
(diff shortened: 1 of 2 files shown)


dwayne-hart (approved):

douglatornell:
@dwayne-hart I meant to mark this as draft. Still working on it!

dwayne-hart (dismissed):

nsoontie (commented):
This looks good to me. The only thing is that I think it's outdated with master. @douglatornell Should we pull the changes from master before approving?

nsoontie (approved):

dwayne-hart:
Code has been merged into the mainline.

PR #800, opened 2020-12-09
Update the hash for the configs submodule

dwayne-hart:
In order to ensure that the Ocean Data Map Project is ready to be deployed in production. The configs hash needs to be updated to the latest version on the configurations mainline.


nsoontie (approved):
Thanks Dwayne. I don't see any issues here but I'm recognizing that it is a lot of overhead to update this everytime there is a change on the configurations repo. Are there any options to make that more streamlined? If not, I don't see the need to have an independent reviewer on pull requests of this type.  Thoughts?

dwayne-hart:
@nsoontie, upon reflection. I believe that this change within the repo would not need to go through a review process. However, it should be noted in the build, packaging, and release documentation. Also, I should note it when dropping a tag on the mainline.

PR #806, opened 2020-12-18
Change Class 4 profiles legend to show GIOPS model label from the outset

douglatornell:
## Background
As @nsoontie recently observed, in the initial display of Class 4 profiles the model profiles are labelled as "Model". The profiles are always from the GIOPS class4 dataset, and, if additional model profiles are added to the plot, the "Model" label changes to read "GIOPS". So, this PR changes the legend label to be "GIOPS" from the outset.

This PR also updates the string formatting code in `plotting.class4.Class4Plotter.plot()` to use f-strings as much as possible. In one place an f-string is not feasible without complicating the code, so the update there is to `str.format()`.

Making the model legend consistently name the models shown is an improvement, as is modernizing the code to drop `%s` string formatting.

No.

## Screenshot(s)
Before:
image

After:
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/plotting/class4.py
+++ b/plotting/class4.py
@@ -156,9 +156,7 @@ class Class4Plotter(Plotter):
                 plt.legend(self.ids, loc='best')
 
         plot_label = ""
-        giops_name = gettext("Model")
-        if len(self.additional_model_names) > 0:
-            giops_name = "GIOPS"
+        giops_name = "GIOPS"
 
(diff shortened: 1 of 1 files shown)


dwayne-hart (approved):
Its good to create a standard way of handling the plotting legend.

Issue #664, opened 2020-02-25
Replace"dimension sniffing" with configuration to select model grid class in `open_dataset()`


douglatornell:
**Priority (low, medium, high)**
Priority: medium

**Is your feature request related to a problem? Please describe.**
This is more about improving the clarity of the design intent of the code, and code readability/understandability than it is problem per se.

**Describe the solution you'd like**
The model grid class to be used for a dataset should be identified in `oceannavigator/datasetconfig.json` so that it is available on `DatasetConfig`. Then `open_dataset()` can use that to determine which model grid class to return.

I think there is also an argument to be made for some renaming of classes to make the design intent and architecture clearer (though, of course, "naming things is hard"). I suggest:
* `Model` becomes `ModelGrid`
* `Mercator` becomes `LatLonModel`, `LatLonGrid`, or `LatLonModelGrid`
* `Nemo` becomes `XYModel`, `IndexGrid`, or `IndexModelGrid`
* `Fvcom` becomes `UnstructuredModel`, `UnstructuredGrid`, or `UnstructuredModelGrid`, but all of those feel too long to me :-(

**Additional context**

re: https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/1cf39c31ed787216626dd06c98829a271e4dcb66/data/__init__.py#L77

PR #811, opened 2021-01-15
Fix UnboundLocalError netCDFData.dimensions

douglatornell:
## Background
Investigation of https://sentry.io/organizations/dfo-ocean-navigator/issues/2134767923 lead to the realization that a `sqlite3.OperationalError` was being masked by an `UnboundLocalError` in `NetCDFData.dimensions`. This PR increases unit test coverage for `dimensions`, and make its result of an empty list consistent for the cases of `OperationsError` and `FileNotFoundError`.

`UnboundLocalError` is a really bad code-smell. At least we now have a consistent outcome when there is a problem accessing sqlite3 and netCDF4 datasets.

The `OperationalError` is likely and indication of a deeper file system or container issue that is causing problems with access to the sqlite3 databases.

It is weird that `dimensions` is opening datasets at all. That should be the exclusive job of `NetCDFData.__enter__()`. We're stuck with the present implementation because `data.open_dataset()` accesses `dimensions` before the `NetCDFData` constructor is called do that it can sniff out which flavour of model dataset to create; see issue #664.

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -83,7 +83,7 @@ class NetCDFData(Data):
                     url = self.url if isinstance(self.url, list) else [self.url]
                     # This will raise a FutureWarning for xarray>=0.12.2.
                     # That warning should be resolvable by changing to:
-                    # fields = xarray.open_mfdataset(self.url, combine="by_coords", decode_times=decode_times)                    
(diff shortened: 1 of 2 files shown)


dwayne-hart (approved):
Glad to see that this embedded bug resolved.

PR #812, opened 2021-02-01
Nuke stats tab

htmlboss:
## Background

Statistics functionality has been broken on the back end for a very long time, with not timeline on it being fixed. Let's just remove the ability for users to request it to reduce confusion.

It's the only way

Nope

## Screenshot(s)
Tab is gone:
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


dwayne-hart (approved):
Nice.

Issue #17, opened 2017-07-13
Determine desired functionality of side-by-side.


htmlboss:
Since this is a very large implementation, we need to make a list of the desired features + fixes so I can prioritize which one to tackle first.

PR #814, opened 2021-02-02
Get model class from config instead of "sniffing" dataset dimensions

### LINKED ISSUE #664: Replace"dimension sniffing" with configuration to select model grid class in `open_dataset()`

douglatornell:
## Background

Resolves #664 

This PR addresses issue #664. It changes the way we decide what model data class to instantiate in `data.__init__()`. Instead of opening a dataset to inspect its dimensions and decided from them what model data class to use, we read the model data class name from `datasetconfig.json`.

* Inspecting the dimensions required us to open the dataset before we really opened the dataset  - a bit convoluted...
* It was not robust with respect to datasets from models that hadn't yet been coded for in `data.__init__()`; e.g. ROMS, perhaps.
* It lead to the "concealed" error in issue #811.

* Making the model data class an explicit attribute of each dataset in `datasetconfig.json` means that the impact of adding a dataset that requires a new model class will be immediately obvious.
* The change leads to simplification of code in several places, in particular in the `NetCDFData.dimensions` property that was the cause of issue #811.
* We are also able to do away with the special case `meta_only` arg and model data class attribute.

Changes to `datasetconfig.json` in the [DFO-Ocean-Navigator/configurations](https://github.com/DFO-Ocean-Navigator/configurations) repo are required in tandem with this PR; please see configurations repo [PR #17](https://github.com/DFO-Ocean-Navigator/configurations/pull/17).

## Screenshot(s)
N/A

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


routes/api_v1_0.py
```diff
@@ -802,10 +778,7 @@ def topo_v1_0(shaded_relief: str, projection: str, zoom: int, x: int, y: int):
         Generates topographical tiles
     """
 
-    if shaded_relief == "true":
-        bShaded_relief = True
-    else:
-        bShaded_relief = False
+    bShaded_relief = shaded_relief == "true"
```
htmlboss, line 781: Nice

htmlboss (commented):
Overall looks good. Def try this on a test container before merge given the changes to some critical paths.

Unrelated note I think we should look at the strategy pattern for figuring out how to open a dataset, given that now we have three possible ways: sqlite, zarr, and dap.

routes/api_v1_0.py
douglatornell: Thanks, @htmlboss. I've been running this code in my dev container for the past couple of weeks. So, I'm confident enough in it to move it to staging for some other folks to exercise.

I agree that it would be nice to have a more uniform interface for opening sqlite, zarr, and dap. Will think on it...

htmlboss (approved):

douglatornell:
@dwayne-hart Nabil has approved this, so when you have time I think it is ready for merge and deployment to staging so that it gets some exercise.

PR #815, opened 2021-02-03
Cast Class 4 ID to S8 when reading from file

htmlboss:
## Background
While checking Greg's class 4 file, I needed to add this cast in order for it to be read, since the files present on Dory have this encoding and his were just a byte array.

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/class4.py
+++ b/data/class4.py
@@ -139,7 +139,7 @@ def class4(class4_id, projection, resolution, extent):
     with xr.open_dataset(dataset_url) as ds:
         lat_in = ds['latitude'][:]
         lon_in = ds['longitude'][:]
-        ids = np.char.decode(ds['id'][:].values, 'UTF-8')
+        ids = np.char.decode(ds['id'].astype('|S8').values, 'UTF-8')
         for i in range(0, lat_in.shape[0]):
             x, y = proj(lon_in[i], lat_in[i])
             p = Point(y, x)


htmlboss:
Should be tested on staging first.

douglatornell (commented):
This took me on a deeper dive than I expected into NumPy dtypes. `S` is for backward compatibility for Python2 zero-terminated byte strings. Do we really need that?

When I test against files like `/data/class4/2020/20201215/class4_20201215_GIOPS_CONCEPTS_3.0_profile.nc` I find that this like can be simplified to:
```python
ids = np.char.decode(ds.id)
```
i.e.
```python
numpy.testing.assert_array_equal(
    np.char.decode(ds['id'].astype('|S8').values, 'UTF-8'), 
    np.char.decode(ds.id)
)
```
passes, and the `routes.api_v1_0.query_file_v1_0()` calls that lead to `data.class4.class4()` complete without error.

I'm really curious to see an example of the new class4 file example that you have from Greg that makes the `S8` casting necessary.

htmlboss:
Here's an example file: ftp://depot.cmc.ec.gc.ca/upload/mrb/greg.smith/20190327_SAM_class4.nc

douglatornell:
> ftp://depot.cmc.ec.gc.ca/upload/mrb/greg.smith/20190327_SAM_class4.nc

Hmmm... I can't seem to access this. Can do anonymous ftp and navigate the directories, but get on the file fails, as does ls at any level in the dir tree. Have emailed Greg.

dwayne-hart:
The file is no longer available.

douglatornell:
I finally got back to this! Sorry for taking so long @htmlboss.

I got a copy of the `2019032700_SAM_class4.nc` file from Greg. `ncdump` shows that the difference between it and the class4 files presently in our collection like `class4_20200101_GIOPS_CONCEPTS_3.0_profile.nc`. In `2019032700_SAM_class4.nc` the variable `id` is of type `string`:
```netcdf \2019032700_SAM_class4 {
dimensions:
        numobs = 1331 ;
        numdeps = 1774 ;
        numvars = 2 ;
        string_length8 = 8 ;
variables:
...
        string id(numobs) ;
                string id:_FillValue = "99999" ;
                id:long_name = "Observation id" ;
                id:_Storage = "contiguous" ;
...
```
while in `class4_20200101_GIOPS_CONCEPTS_3.0_profile.nc` it is of type `char`:
```
netcdf class4_20200101_GIOPS_CONCEPTS_3.0_profile {
dimensions:
        numobs = 660 ;
        numvars = 2 ;
        numdeps = 1306 ;
        string_length8 = 8 ;
        string_length128 = 128 ;
        numfcsts = 10 ;
variables:
...
        char id(numobs, string_length8) ;
                id:long_name = "Observation id" ;
...
```
`char` gets interpreted as an array of groups of bytes and has to be decoded to get an array of strings, while `string` is just that - an array of strings.

Trying to decode a string raises a `TypeError` exception. So, I think the modern, Python3 and xarray way to write lines 140-142 to handle old and new class4 files is:
```python
        lat_in = ds.latitude
        lon_in = ds.longitude
        try:
            ids = np.char.decode(ds.id)
        except TypeError:
            ids = ds.id.data
```

JustinElms:
I'm not sure if this is still necessary, Closing until a related issue comes up.

PR #817, opened 2021-02-19
Add selectbox for choosing bathymetry layer source

htmlboss:
## Background

Ignore the code formatting changes, it was automatically done by the VS code linter lol

The following bathymetry contour layers will be available:
* ETOPO1

* Did some basic refactoring (function extraction) to make this spaghetti code a little more digestible lmao.

* Started implementing this `shouldUpdate` framework to gradually improve perceived responsiveness of website. As more of these things are added, the number of components updated per react update cycle will reduce and thus will happen quicker. It's just standard react best practices that we haven't gotten around to implementing yet.

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map.jsx
@@ -26,14 +26,14 @@ const X_IMAGE = require("../images/x.png");
 
 var app = {};
 const COLORS = [
-  [ 0, 0, 255 ],
-  [ 0, 128, 0 ],
-  [ 255, 0, 0 ],
-  [ 0, 255, 255 ],
-  [ 255, 0, 255 ],
-  [ 255, 255, 0 ],
-  [ 0, 0, 0 ],
-  [ 255, 255, 255 ],
+  [0, 0, 255],
+  [0, 128, 0],
+  [255, 0, 0],
+  [0, 255, 255],
+  [255, 0, 255],
+  [255, 255, 0],
+  [0, 0, 0],
+  [255, 255, 255],
 ];
 
 let CURRENT_PROJ = "EPSG:3857";
@@ -65,7 +65,7 @@ const MAX_ZOOM = {
 var drifter_color = {};
 
 // Reset Pan button
-app.ResetPanButton = function(opt_options) {
+app.ResetPanButton = function (opt_options) {
   const options = opt_options || {};
 
   const button = document.createElement("button");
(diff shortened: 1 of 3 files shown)


oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -77,6 +77,7 @@ export default class OceanNavigator extends React.Component {
         mapBathymetryOpacity: 0.75, // Opacity of bathymetry contours
         topoShadedRelief: false,    // Show hill shading (relief mapping) on topography
         bathymetry: true,           // Show bathymetry contours
+        bathyContour: 'etopo1',
```
htmlboss, line 80: default to ETOPO1

htmlboss:
So it seems the CHS bathymetry isn't transparent like our ETOPO is...
image

This could be problematic

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -1174,21 +1174,83 @@ export default class Map extends React.PureComponent {
     this.layer_obsDraw.setZIndex(6);
   }
 
+  shouldUpdateScaleViewer(currentDataset, prevDataset,
```
htmlboss, line 1177: All the changes above this line are just auto-formatting

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -1202,18 +1264,25 @@ export default class Map extends React.PureComponent {
 
     datalayer.setSource(newSource);
 
-    // Update colour scale
-    if (this.scaleViewer != null) {
-      this.map.removeControl(this.scaleViewer);
+    if (this.shouldUpdateScaleViewer(this.props.state.dataset,
```
htmlboss, line 1267: Added a shouldUpdate check here since otherwise we'd be reconstructing the variable scale bar every time the component updates which is silly. We only need to reconstruct when the dataset, variable, or actual variable scale values change (0, 30 for example).

Checked it and it works fine...it does disappear for a little bit when changing variables or datasets but that's a result of crappy javascript architecture.

dwayne-hart:
Nice, I was able to get this integrated and working with the staging site.

image

dwayne-hart (approved):
Nice bit of code cleanup.

PR #821, opened 2021-03-17
Issue/784 816 fixing the start date for vm

BaruaSourav:
## Background
When VM tab is selected upon selecting a point, the Point Window component was showing the loading animation and the start time is loaded as NaN. And also in the backend it shows **IndexError: index 0 is out of bounds for axis 0 with size 0** as seen in [Issue 816](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/816). There was a flaw in the calculation of the _starttime_ state variable in _PointWindow.jsx:Line47_, which was causing _netcdf_data.timestamp_to_time_index()_ to face an indexerror. 
Instead of calculating the starttime state variable from props.time, we can use the props.starttime as the starttime for PointWindow component. We can pass the starttime from the parent OceanNavigator component as the props for PointWindow (Which we are already doing, but not trying to access that from PointWindow component).

On each time we change the dataset, and call changeDataset() method OceanNavigator.jsx: Line 257, we can change the global state.starttime to id of first element of the result from the call to /api/v1.0/timestamps/?dataset=[DATASET]&variable=[VARIABLE]. This call returns an array of the available timestamps and setting starttime to the first element's id of this returned array, sets the starttime to the earliest available timestamp on that selected dataset. This starttime value is then passed to the PointWindow component and can be used to perform the starttime calculation on various context, including the VM plot.

Although, this method will set the starttime value when user changed the dataset once (As the code steps in the changeDataset() function after user changes a dataset). So, I have also set the OceanNavigator component so that, it gets the timestamps for "giops_day" dataset and "votemper" variable as it is the default first selected dataset and variable. 

But for long time-series datasets like SalishSeaCast (hourly since 2007-01-01, so >100,000 timestamps). Setting _starttime_ to its first available timestamp would result in a large amount of backend task, so the _starttime_ is set to first element of the timestamp array if the length of timestamp array is less than 24. If it is larger than 24, it is set as the 24th element from the end of the timestamp array. 

## Checks
- [x] I ran unit tests.
I checked different datasets' VM tab after selecting a point on the global map, and the start time is properly set in my developer environment.

### DIFF
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -123,6 +123,15 @@ export default class OceanNavigator extends React.Component {
     this.updateScale = this.updateScale.bind(this);
     this.get_variables_promise = this.get_variables_promise.bind(this);
     this.get_timestamp_promise = this.get_timestamp_promise.bind(this);
+    // Setting the starttime and time variable default value
+    const time_promise = this.get_timestamp_promise("giops_day", "votemper");
+      $.when(time_promise).done(function(time) {
+        let newTime = time[time.length-1].id;
+        let newStarttime = time[0].id;
+        this.state.time = newTime;
+        this.state.starttime = newStarttime;
+        this.setState(state);
+      }.bind(this));
   }
   
   //Updates the page language upon user request
@@ -272,16 +281,18 @@ export default class OceanNavigator extends React.Component {
 
 
       let newTime = 0;
+      let newStarttime = 0;
       const time_promise = this.get_timestamp_promise(dataset, newVariable);
       $.when(time_promise).done(function(time) {
         newTime = time[time.length-1].id;
-
+        newStarttime = time[0].id;
         if (state === undefined) {
           state = {};
         }
         state.dataset = dataset;
         state.variable = newVariable;
         state.time = newTime;
+        state.starttime = newStarttime;
         state.busy = false;
 
         this.setState(state);
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -285,7 +285,11 @@ export default class OceanNavigator extends React.Component {
       const time_promise = this.get_timestamp_promise(dataset, newVariable);
       $.when(time_promise).done(function(time) {
         newTime = time[time.length-1].id;
-        newStarttime = time[0].id;
+        // check the length of the timestamp array and set the start time according to that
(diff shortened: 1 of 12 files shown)


oceannavigator/frontend/src/components/PointWindow.jsx
```diff
@@ -44,14 +44,17 @@ export default class PointWindow extends React.Component {
...
+      starttime: props.starttime,
       variables: [],
       variable: [props.variable],
       observation_variable: [0],
       size: "10x7",
       dpi: 144,
       plotTitles: Array(7).fill(""),
     };
+    //console.log("State [Starttime] "+this.state.starttime);
+    //console.log("Props [Time] "+ props.starttime);
```
htmlboss, line 57:
```suggestion
```

oceannavigator/frontend/src/components/PointWindow.jsx
```diff
@@ -44,14 +44,17 @@ export default class PointWindow extends React.Component {
       depth: props.depth,
       showmap: false,
       colormap: "default",
-      starttime: Math.max(props.time - 24, 0),
+      //starttime: Math.max(props.time - 24, 0),
```
htmlboss, line 47:
```suggestion
```

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -272,16 +279,22 @@ export default class OceanNavigator extends React.Component {
 
 
       let newTime = 0;
+      let newStarttime = 0;
       const time_promise = this.get_timestamp_promise(dataset, newVariable);
       $.when(time_promise).done(function(time) {
         newTime = time[time.length-1].id;
-
+        // check the length of the timestamp array and set the start time according to that
```
htmlboss, line 286:
```suggestion
```

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -272,16 +279,22 @@ export default class OceanNavigator extends React.Component {
...
+        if (time.length > 24)
+          newStarttime = time[time.length-25].id;
+        else
+          newStarttime = time[0].id;
```
htmlboss, line 290: Could we move this bit of logic into its own function? Maybe something like `getStartTime`...The calculation looks fine to me though!

htmlboss (changes requested):
Few comments/suggestions for changes. Also it looks like a rebase went bad since you have a bunch of merged changes? After rebasing off master I usually do a `git push --force-with-lease`.

PR #808, opened 2021-01-03
Re-write data API endpoint

htmlboss:
## Checks
- [ ] I ran unit tests.
- [ ] I've tested the relevant changes from a user POV.

_Hint_ To run all python unit tests run the following command from the root repository directory:
```sh
bash run_python_tests.sh
```

PR #822, opened 2021-03-19
Add GeoJSON transformer to data package

htmlboss:
## Background

Prerequisite for getting the `/data` API endpoint back up and running (#808)

* Provides functionality to serialize an xarray data array into a valid geojson dictionary.
* Added new `trunc` helper function to correctly truncate floating-point values to a given number of decimal places.
* Added relevant tests.

I decided to implement this as a free function to start instead of dumping it into `Data` as a virtual function for the sake of faster iteration....just wanna get #808 out before end of fiscal.

Ultimately I'd like to see syntax like `my_mercator_instance[0, 0, 5:10, 5:10].to_geojson()`, `...to_netcdf(nc_file_type)`, `...to_csv()`, etc... but `Data` is also missing `__getitem__` for proper slicing (so we can ditch `get_point`, `get_area`, etc, and simply keep the helper functions like `get_line`, `get_profile).

On the other hand, this free-function style could stay and allow us to keep `Data` subclasses lean....up for discussion another time.

Nope.

## Screenshot(s)
n/a

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/transformers/geojson.py
+++ b/data/transformers/geojson.py
@@ -1,0 +1,56 @@
+import numpy as np
+import xarray as xr
+from data.utils import trunc
+
+from geojson import Feature, FeatureCollection, Point
+
+
+class WrongNumberOfDimensionsError(ValueError):
+    pass
+
+
+def data_array_to_geojson(data_array: xr.DataArray, lat_key: str, lon_key: str) -> FeatureCollection:
+    """
+    Converts a given xarray.DataArray, along with lat and lon keys to a geojson.FeatureCollection (subclass of dict).
+    
+    A FeatureCollection is really just a list of geojson.Feature classes.
+    Each Feature contains a geojson.Point, and a `properties` dict which holds arbitrary attributes
+    of interest for a Point. In the case of this function, each Feature has the following properties:
+
+    * The data value corresponding to a lat/lon pair (e.g. salinity or temperature)
+    * A copy of the `attribs` field held in the data_array (yes there's lots of duplication...blame the geojson spec).
+
+
+    Important notes:
+
+    * All data values are truncated to 3 decimal places.
+
+    Parameters:
+        * data_array -- A 2D field (i.e. lat/lon only...time and depth dims should be sliced out).
+        * lat_key -- Key of the latitude coordinate (e.g. "latitude").
+        * lon_key -- Key of the longitude coordinate (e.g. "longitude").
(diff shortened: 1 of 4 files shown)


douglatornell:
I'm guessing that the the `geojson` package you had to add to your dev env for this is https://pypi.org/project/geojson/ correct?

dwayne-hart:
Hi @douglatornell, this package was requested by @htmlboss a little while ago. I had installed version 2.5.0 from conda-forge. As well, I had created a new spec file for the navigator toolchain.

```
(navigator) ubuntu@u2004-staging:~$ conda list | grep geojson
geojson                   2.5.0                      py_0    conda-forge
```

douglatornell:
As requested, @htmlboss, I've done some Pythonic buffing. Assumed you'd be okay with me just pushing into the PR branch rather than telling you what to type (and proud of myself for figuring out how to do that :grin: )

I changed to a `ValueError` with a custom message instead of a new custom `WrongNumberOfDimensionsError` exception, but if you have a use case in mind for the custom exception, apologies - I'll change it back. Moved the check for that exception up in the code - fail-fast principle.

A bunch of changes to use the `xarray` object model rather than converting to a `dict`. Doing that lead me to the interesting quirk about values from netCDF4 datasets (at least the `testdata/giops_test.nc`) coming in as 32-bit floats, and `json` only supporting serialization of 64-bit floats. That seemed worth an explanatory comment in the code to avoid future-dev wasting time on trying to get rid of a seemingly superfluous type cast.

Changed calculation of `Point()` collection to a generator expression because we don't need it as a list, just need the points as a stream.

Simplified list comprehension iteration to produce list of `Feature()`.

Of course you having written unit tests made it sooo much easier for me to do all of this! :tada:

htmlboss:
Thanks for the revisions! Things are much simpler and more lazily evaluated (generator expression) which should help Dory a fair bit. Now that I've [rtfm'ed](https://docs.python.org/3/howto/functional.html#generator-expressions-and-list-comprehensions), I wonder if there are other spots where we could benefit from lazy evaluation :joy:

Not sure why I converted the dataarray to a dict at first 🤷  midnight oil must have been running low at the time.

@dwayne-hart this is ready for review

dwayne-hart (approved):
Looks clean and tidy.

PR #824, opened 2021-03-23
Issue/784 816 fixing start date of vm

BaruaSourav:
## Background
When VM tab is selected upon selecting a point, the Point Window component was showing the loading animation and the start time is loaded as NaN. And also in the backend it shows **IndexError: index 0 is out of bounds for axis 0 with size 0** as seen in [Issue 816](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/816). There was a flaw in the calculation of the _starttime_ state variable in _PointWindow.jsx:Line47_, which was causing _netcdf_data.timestamp_to_time_index()_ to face an indexerror. 
Instead of calculating the starttime state variable from props.time, we can use the props.starttime as the starttime for PointWindow component. We can pass the starttime from the parent OceanNavigator component as the props for PointWindow (Which we are already doing, but not trying to access that from PointWindow component).

On each time we change the dataset, and call changeDataset() method OceanNavigator.jsx: Line 257, we can change the global state.starttime to id of first element of the result from the call to /api/v1.0/timestamps/?dataset=[DATASET]&variable=[VARIABLE]. This call returns an array of the available timestamps and setting starttime to the first element's id of this returned array, sets the starttime to the earliest available timestamp on that selected dataset. This starttime value is then passed to the PointWindow component and can be used to perform the starttime calculation on various context, including the VM plot.

Although, this method will set the starttime value when user changed the dataset once (As the code steps in the changeDataset() function after user changes a dataset). So, I have also set the OceanNavigator component so that, it gets the timestamps for "giops_day" dataset and "votemper" variable as it is the default first selected dataset and variable. 

But for long time-series datasets like SalishSeaCast (hourly since 2007-01-01, so >100,000 timestamps). Setting _starttime_ to its first available timestamp would result in a large amount of backend task, so the _starttime_ is set to first element of the timestamp array if the length of timestamp array is less than 24. If it is larger than 24, it is set as the 24th element from the end of the timestamp array. 

## Checks
- [x] I ran unit tests.
I checked different datasets' VM tab after selecting a point on the global map, and the start time is properly set in my developer environment.

### DIFF
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -124,15 +124,6 @@ export default class OceanNavigator extends React.Component {
     this.get_variables_promise = this.get_variables_promise.bind(this);
     this.get_timestamp_promise = this.get_timestamp_promise.bind(this);
     this.setStartTime = this.setStartTime.bind(this);
-    // Setting the starttime and time variable default value
-    const time_promise = this.get_timestamp_promise("giops_day", "votemper");
-      $.when(time_promise).done(function(time) {
-        let newTime = time[time.length-1].id;
-        let newStarttime = time[0].id;
-        this.state.time = newTime;
-        this.state.starttime = newStarttime;
-        this.setState(state);
-      }.bind(this));
   }
   
   //Updates the page language upon user request
@@ -524,6 +515,15 @@ export default class OceanNavigator extends React.Component {
     } else {
       window.history.back();
     }
+  }
+  componentDidMount(){
+    // Setting the starttime and time variable default value
+    const time_promise = this.get_timestamp_promise("giops_day", "votemper");
+      $.when(time_promise).done(function(time) {
+        const newTime = time[time.length-1].id;
+        const newStarttime = time[0].id;
+        this.setState({time: newTime, starttime: newStarttime});
+      }.bind(this));
   }
 
   componentDidUpdate(prevProps, prevState) {
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -123,6 +123,15 @@ export default class OceanNavigator extends React.Component {
     this.updateScale = this.updateScale.bind(this);
     this.get_variables_promise = this.get_variables_promise.bind(this);
     this.get_timestamp_promise = this.get_timestamp_promise.bind(this);
+    // Setting the starttime and time variable default value
+    const time_promise = this.get_timestamp_promise("giops_day", "votemper");
+      $.when(time_promise).done(function(time) {
+        let newTime = time[time.length-1].id;
+        let newStarttime = time[0].id;
+        this.state.time = newTime;
+        this.state.starttime = newStarttime;
+        this.setState(state);
+      }.bind(this));
   }
   
   //Updates the page language upon user request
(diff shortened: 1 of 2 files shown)


oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -123,6 +123,16 @@ export default class OceanNavigator extends React.Component {
...
+    this.setStartTime = this.setStartTime.bind(this);
+    // Setting the starttime and time variable default value
+    const time_promise = this.get_timestamp_promise("giops_day", "votemper");
+      $.when(time_promise).done(function(time) {
+        let newTime = time[time.length-1].id;
+        let newStarttime = time[0].id;
+        this.state.time = newTime;
+        this.state.starttime = newStarttime;
+        this.setState(state);
+      }.bind(this));
```
htmlboss, line 135: I'm a bit concerned with firing a promise in the constructor of a React object, since `setState` will cause the component to re-render a bunch of times....

could this `time_promise` be put inside a `componentDidMount()`?

See docs: https://reactjs.org/docs/react-component.html#componentdidmount

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -123,6 +123,16 @@ export default class OceanNavigator extends React.Component {
...
     this.get_timestamp_promise = this.get_timestamp_promise.bind(this);
+    this.setStartTime = this.setStartTime.bind(this);
+    // Setting the starttime and time variable default value
+    const time_promise = this.get_timestamp_promise("giops_day", "votemper");
+      $.when(time_promise).done(function(time) {
+        let newTime = time[time.length-1].id;
+        let newStarttime = time[0].id;
+        this.state.time = newTime;
+        this.state.starttime = newStarttime;
+        this.setState(state);
```
htmlboss, line 134: With the call to `this.setState(state);`, haa the `state` variable been defined?

htmlboss, line 134: What do you think of simplifying to the following? Also, you should avoid "direct state mutation" (i.e. use `this.setState` over `this.state.myVariable = `. This is because React does a bunch of async batched updates under the hood with `setState` to improve performance.
```suggestion
        const newTime = time[time.length-1].id;
        const newStarttime = time[0].id;
        this.setState({time: newTime, starttime: newStarttime});
```

htmlboss (changes requested):
Super happy to see this finally fixed! Just a couple of comments regarding React-specific stuff...

oceannavigator/frontend/src/components/OceanNavigator.jsx
BaruaSourav: Moved the `time_promise` promise to the` componentDidMount()` lifecycle hook.

oceannavigator/frontend/src/components/OceanNavigator.jsx
BaruaSourav: Changed the state using `this.setState()` instead of directly mutating the state.

BaruaSourav: Used `this.setState()` instead of the direct state mutation.

douglatornell (approved):
Tested this in my dev env for both GIOPS 10d forecast and SalishSeaCast T&S datasets and it makes me happy! :grinning: 

Updating response of various widgets in the plot window after each change is annoyingly slow, but that's a separate issue for another day. These changes resolve the issue I raised from Sentry in #816.

dwayne-hart (approved):
Seeing that @BaruaSourav addressed the points that @htmlboss raised and successful testing by @douglatornell. This work will be merged in.

dwayne-hart (approved):

dwayne-hart (approved):

PR #825, opened 2021-03-24
Removed references to 'scale_factor'

### LINKED ISSUE #729: Deprecate `scale_factor` in datasetconfig.json in favour of equation
The calculation layer makes the use of the `scale_factor` attribute in the datasetconfig.json redundant, since an `equation` can easily replicate the functionality: 
`scale_factor: 1.2` can be replaced by
`equation: "my_variable * 1.2", dims: [dims...]`.

The only work needed is editing the config file. The calculation layer handles the rest automatically. 

The other (optional but strongly encouraged) part is removing all references/conditionals related to scale_factor in the Python code. Places to look are in the plotting classes (`load_data` functions specifically) and in the `DatasetConfig` class.

Overall this will slightly reduce the total Python code we have while making use of newer and more efficient code paths.

JustinElms:
## Background
A 'scale_factor' attribute is given for many variables within the datasetconfig.json file which scales the data as needed for the presentation layer. While an important function it can be replaced by an equation which will provide the same functionality with more efficiency. 

References to 'scale_factor' were removed from datasetconfig.json, dataset_confi.py, and several plotting classes. In some instances where the scale factor was applied to the data along with other operations. For these cases the code was appropriately modified to preserve the necessary operations. 

closes #729 
 
Though 'scale_factor' can be depreciated just by removing references in datasetconfig.json, references in the calculation layer were also removed to help reduce the total amount of code in the Navigator. 

## Checks
- [x] I ran unit tests.

### DIFF
--- a/oceannavigator/dataset_config.py
+++ b/oceannavigator/dataset_config.py
@@ -326,22 +326,6 @@ class VariableConfig:
             return [0, 100]
 
     @property
-    def scale_factor(self) -> float:
-        """
-        Returns variable scale factor from dataset config file
-        """
-
-        try:
-            scale_factor = self.__get_attribute("scale_factor")
-        except KeyError:
-            scale_factor = None
-
-        if scale_factor is not None:
-            return scale_factor
-        else:
-            return 1.0
-
-    @property
     def is_hidden(self) -> bool:
         """
         Is the given variable marked as hidden in the dataset config file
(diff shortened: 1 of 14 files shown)


douglatornell:
Woohoo! 100 fewer lines of code to maintain! Thanks for going the extra step of finding and deleting all the code that the config change made obsolete.

dwayne-hart (approved):
Because of @douglatornell 's suggestion this work is being merged into the mainline.

PR #829, opened 2021-04-02
Re-write /data endpoint

htmlboss:
## Background

First cut at re-writing the /data endpoint to return standard geojson so that it's easier for consumers to use our data. This will be the first endpoint of ours that will be "protected" by a schema definition. Over time all of our endpoints should have one to reduce the validation boilerplate.

Usage `GET /api/v1.0/data?...`

The endpoint now requires the following parameters:
* dataset
* variable
* time
* depth
* geometry_type 

The following meta data is returned for each lat/lon coord:
* Variable value
* Short name
* Long name
* Units

* Added relevant tests.

`parsetab.py` and `parser.out` shouldn't be checked into git....should probably to gitignore.

Used `geojson` package to do the heavy lifting of serialization and object construction. `marshmallow` enforces the api schema and handles all of the error messaging and what not. We just send it back to the client with a `400` status code.

Also had to work around the nonsense of serializing numpy data types into json.

Took a slightly different approach with placing the test files, but I think it should make them more digestible in the longer run as opposed to lopping every file in one folder.

Specifics are in review comments.

Only that 1.5 million data points can be returned in under 2 secs on my local machine which uses a long-distance connection to the data server...so this number should be way lower on the production container. Payload size for said points is under 10kb which is very acceptable to me.

## Screenshot(s)
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


routes/api_v1_0.py
```diff
@@ -301,31 +306,31 @@ def range_query_v1_0(dataset: str, variable: str, interp: str, radius: int, neig
...
-            dataset, variable, date, depth,
-            list(map(float, location.split(",")))
-        )
-        resp = jsonify(data)
-        resp.cache_control.max_age = 2
-        return resp
+    try:
+        result = GetDataSchema().load(request.args)
+    except ValidationError as e:
+        print(e)
```
htmlboss, line 318: nuke this print

routes/api_v1_0.py
```diff
@@ -301,31 +306,31 @@ def range_query_v1_0(dataset: str, variable: str, interp: str, radius: int, neig
...
+        abort(400, str(e))
+
+    config = DatasetConfig(result['dataset'])
+    
+    with open_dataset(config, variable=result['variable'], timestamp=result['time']) as ds:
+
+        print(ds.nc_data.get_dataset_variable(result['variable']).shape)
```
htmlboss, line 325: nuke this print

data/transformers/geojson.py
```diff
@@ -33,10 +33,16 @@ def data_array_to_geojson(data_array: xr.DataArray, lat_key: str, lon_key: str)
...
+
     # Need to ensure that data values are 64-bit floats (i.e. Python builtin float) because
     # that's the only type of float that json will serialize without a custom serializer.
     # Floats from netCDF4 datasets are often 32-bit.
-    data = trunc(data_array.data.astype(float))
+    data = np.nan_to_num(
+        trunc(
+            data_array.astype(float)
+        ),
+        nan=np.Infinity
```
htmlboss, line 44: nans can't be serialized into json natively so filter them out with a standard constant.

data/calculated.py
```diff
@@ -121,13 +121,13 @@ def __init__(self, parent, expression, dims, attrs={}, db_url=""):
...
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: str) -> xr.DataArray:
         # This is where the magic happens.
 
         data_array = self._parser.parse(
             self._expression, self._parent, key, self._dims)
 
-        return xr.DataArray(data_array)
+        return xr.DataArray(data_array, attrs=self.attrs)
```
htmlboss, line 130: Fixed two year old bug where DataArray generated from calculation layer didn't have the original variable's attributes LOL

JustinElms (commented):
I pulled the `rewrite-data-endpoint` branch and ran it without issue but this test failed on my instance:

`======================================================================
FAIL: test_data_array_to_geojson_builds_correct_feature_collection (tests.data.transformers.test_geojson.GeoJSONTest)
----------------------------------------------------------------------
  File "/home/ubuntu/Ocean-Data-Map-Project/tests/data/transformers/test_geojson.py", line 41, in test_data_array_to_geojson_builds_correct_feature_collection
    self.assertEqual(expected, result)
AssertionError: {'features': [{'geometry': {'coordinates':[979 chars]ion'} != {"features": [{"geometry": {"coordinates":[1045 chars]ion"}

----------------------------------------------------------------------`

Does the `expected` variable need to be updated to match these changes?

Otherwise everything looks great on my end!

htmlboss:
Good catch, thanks! I think something got caught in my stash that should have been committed :joy:

PR #830, opened 2021-04-10
Bugfix: make `get_dataset_variable` return `xarray.DataArray` instead of `xarray.Variable`

htmlboss:
## Background

While working on #829, I ended up discovering a nasty bug I introduced a few years ago (classic), where `CalculatedData.get_dataset_variable()` and `NetCDFData.get_dataset_variable()` were grabbing the wrong representation of the variable in a dataset, by returning a low-level `xarray.Variable`. I discovered this when I tried to do things like `my_variable['latitude']` and I was greeted with a nice old `TypeError: invalid indexer array, does not have integer dtype`. By playing around in `ipython` is when I realized my data types were different (because the `DataArray` allowed me to do `my_variable['latitude']`). Then I looked up the differences between a DataArray and a Variable and found these docs:

image

https://xarray.pydata.org/en/stable/terminology.html#term-DataArray

* Fixed bug.
* Added test coverage in relevant spots.

Followed the xarray docs.

Since this touches how we extract data from our files, I did some extra testing on my local website to make sure no side effects were introduced.

Surprised I didn't encounter this bug sooner.

## Screenshot(s)
Calculated path:
image

Non-calculated path:
image

Random plot:
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -691,11 +691,11 @@ class NetCDFData(Data):
 
         return y_dim, x_dim
 
-    def get_dataset_variable(self, key: str):
+    def get_dataset_variable(self, key: str) -> xarray.DataArray:
         """
-        Returns the value of a given variable name from the dataset
+        Returns the xarray.DataArray for a given variable key
         """
-        return self.dataset.variables[key]
+        return self.dataset[key]
 
     @property
     def variables(self) -> VariableList:
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -59,7 +59,7 @@ class CalculatedData(NetCDFData):
                                    attrs,
                                    self.url)
         else:
-            return self.dataset.variables[key]
+            return super().get_dataset_variable(key)
 
     @property
     def variables(self):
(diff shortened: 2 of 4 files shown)


data/netcdf_data.py
```diff
@@ -691,11 +691,11 @@ def yx_dimensions(self) -> Tuple[str, str]:
 
         return y_dim, x_dim
 
-    def get_dataset_variable(self, key: str):
+    def get_dataset_variable(self, key: str) -> xarray.DataArray:
         """
-        Returns the value of a given variable name from the dataset
+        Returns the xarray.DataArray for a given variable key
         """
-        return self.dataset.variables[key]
```
htmlboss, line 698: Pretty sure I caused this when I migrated from the python `netcdf4` library to `xarray` like 3 years ago :joy_cat:

htmlboss:
I don't think we need to worry about the calculated path of `get_dataset_variable` where it returns a `CalculatedArray`, because `CalculatedArray.__getitem__` returns an `xarray.DataArray` which is the last-minute conversion before we actually need it, so it's fine. If #829 still doesn't work, I'll throw up a follow-up fix.

dwayne-hart (approved):
Nice clean work.

PR #837, opened 2021-05-03
Skip broken kmls

JustinElms:
## Background
If the Navigator encounters an error when parsing a kml file it currently results in the point/line/area menu not being populated. Now that we have a kml file hosted on LFS this is more likely to happen to users if they try to run the Navigator without LFS on their system or before the file is downloaded. Adding a try statement to `list_kml_files` in `misc.py` allows the Navigator to skip kml files when it encounters a parse error and continue to populate the relevant menu.  

Also removed 'Final' from SAR Areas name.

Seemed to be the easiest way to ensure that the Navigator will continue to read in kml files in the event of an error.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/kml/area/SAR_Areas.kml
+++ b/kml/area/SAR_Areas.kml
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:f3718ec2d876b3425159050ce0355a86ab8cea239dec216db4d44a130c134ae3
-size 117738269
+oid sha256:218a9d2371ebb1e5e5b56e4d544b58d6ddf914568603faa0fba361e2dc4480ce
+size 117738263
--- a/utils/misc.py
+++ b/utils/misc.py
@@ -21,10 +21,14 @@ def list_kml_files(subdir):
 
     files = []
     for f in os.listdir(DIR):
+        print(f)
         name = None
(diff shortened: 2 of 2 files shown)


JustinElms:
I don't like the idea of using a Try/Except/Continue but this seems like a case where it could be a clean fix for this problem.  @htmlboss @BaruaSourav is there a better alternative for this issue?

BaruaSourav:
Couldn't think of any other way to achieve this. I think a little improvement would be to change the _except_ clause to catch a specific exception. Do you remember what was the exception that was thrown when parser could not read the KML file? May be a **ParserError** or some other name. If you catch that specific exception object, one of the issues detected by CodeFactor will be removed. 
And, we can also make the error go to the Sentry while it steps into the _exception_, although, as LFS is there at production and staging environment, there will be no errors reported to Sentry.

BaruaSourav (approved):

JustinElms:
Good catch @BaruaSourav, exception type added! 
Sentry should automatically pick up on other errors but I'll to see if I can get it to specifically check for errors in the Try statement.

dwayne-hart (approved):

PR #843, opened 2021-05-21
Add bearing to water velocity timeseries csv and ensure timeseries data is ordered

JustinElms:
## Background
CCG requested that we include water velocity bearing in timeseries csv's. When working on adding this data to the csv output I also found two other problems that are addressed here:

1. The  `Save Image - CSV` function on the Virtual Mooring window gives an Internal Server Error in all instances.
2. The order of point data in the Virtual Mooring timeseries changed each time I restarted my local Navigator instance which lead to concerns about the integrity of VM data. This was also observed when comparing VM plots on the production and staging sites - recreating the same plot on each produced different results - and when changing the start and end times of the timeseries plots. 

It looks like work on including bearing in the water velocity csv output was started in the past so I built on the existing unfinished code pieces to complete the feature. In the `load_data` function  there was a check for 'mag' in the variable name to see if we were plotting water velocity. Here I added functionality to pull the `vozocrtx` and `vomecrty` data using the `get_timeseries_profile` and 
 `get_timeseries_point` functions then adding it as `self.quiver_data`. When the `csv` function is called it checks for this attribute and if present adds the additional "Water East Velocity", "(m/s)", "Water North Velocity", "(m/s)", "Bearing (degrees clockwise positive from North)" columns to the csv. The code for the calculation work was already present so I leveraged that to create the data.

There other issues turned out to be quick fixes. 1 above was resolved by changing `self.depths` to `self.depth` on line 173. Issue 2 was caused by our use of `list(set(self.__flatten_list(file_list)))` to remove duplicates from the list. Apparently this doesn't preserve the list order (TIL!). As of python 3.5 dictionaries maintain their order so this was changed to `list(dict.fromkeys(self.__flatten_list(file_list)))` which appears to produce nc file lists in a consistent order for the tests that I've run.

For the bulk of this work I expanded on a couple areas in the code that appear to be intended for this feature. From there I used our existing functionality to pull the water velocity east and north components so that the bearing could be calculated. 

I'm assuming that the water velocity vector components will always be `vozocrtx` and `vomecrty`, then hard-coding the additional csv headers as "Water East Velocity (m/s)" and "Water North Velocity (m/s)". This doesn't feel like a good practice but I didn't see a clean way to pull that data from somewhere else. If there's a better way to do this let me know.

## Checks
- [X] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/plotting/timeseries.py
+++ b/plotting/timeseries.py
@@ -108,9 +108,9 @@ class TimeseriesPlotter(PointPlotter):
             
             if 'mag' in variable and self.depth != 'all':
                 # Under the current API this indicates that velocity data is being
-                # loaded. Save each velocity component (X and Y) for possible CSV
-                # export later. Only provides velocity components for a single
-                # depth. 
+                # loaded. Save each velocity vectorcomponent (X and Y) for possible 
+                # CSV export later. Currently, we only provide velocity components 
+                # for a single depth. 
 
                 vector_variables = [
                     self.dataset_config.vector_variables[variable]['east_vector_component'],
@@ -164,7 +164,7 @@ class TimeseriesPlotter(PointPlotter):
         have_quiver = hasattr(self, 'quiver_data')
 
         if self.depth != 'all':
-            if isinstance(self.depth, str) or isinstance(self.depth, str):
+            if isinstance(self.depth, str):
                 header.append(["Depth", self.depth])
             else:
                 header.append(
--- a/plotting/timeseries.py
+++ b/plotting/timeseries.py
@@ -119,7 +119,7 @@ class TimeseriesPlotter(PointPlotter):
                     for p in self.points:
                         velocity_data = []
                         if self.depth == 'all':
-                            d, depths = dataset.get_timeseries_profile(
+                            d, velocity_depths = dataset.get_timeseries_profile(
                                 float(p[0]),
                                 float(p[1]),
                                 self.starttime,
@@ -141,16 +141,6 @@ class TimeseriesPlotter(PointPlotter):
 
                     self.quiver_data = velocity_point_data
 
-                """
-                self.quiver_data = [point_data[:, 0, :], point_data[:, 1, :]]
-
-                point_data = np.ma.expand_dims(
-                    np.sqrt(
-                        point_data[:, 0, :] ** 2 + point_data[:, 1, :] ** 2
-                    ), 1
-                )
-                """
-
             self.times = times
             self.data = point_data
             self.depths = depths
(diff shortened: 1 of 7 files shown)


plotting/timeseries.py
```diff
@@ -110,15 +110,36 @@ def load_data(self):
...
-                """
-                self.quiver_data = [point_data[:, 0, :], point_data[:, 1, :]]
 
-                point_data = np.ma.expand_dims(
-                    np.sqrt(
-                        point_data[:, 0, :] ** 2 + point_data[:, 1, :] ** 2
-                    ), 1
-                )
-                """
+                velocity_variables = ['vozocrtx', 'vomecrty']
```
htmlboss, line 114: We actually have other velocity variables like sea ice and surface wind. So the problem turns into how do we find the component variables when given a magnitude variable? Our magnitude variables always start with `mag` so that might be something to consider... This will also affect how the CSV column headers are set below (i.e. we can't have them hard-coded to water velocity)

plotting/timeseries.py
```diff
@@ -137,34 +158,28 @@ def csv(self):
...
-        else:
-            have_quiver = False
 
         if self.depth != 'all':
             if isinstance(self.depth, str) or isinstance(self.depth, str):
                 header.append(["Depth", self.depth])
             else:
                 header.append(
-                    ["Depth", "%.4f%s" % (self.depths,
+                    ["Depth", "%.4f%s" % (self.depth,
```
htmlboss, line 173: By changing this to `depth`, how would this affect a timeseries CSV over all depths? See https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/master/tests/test_api_v_1_0.py#L296

htmlboss: DOes it make sense to create a CSV for the whole water column?

JustinElms:
@htmlboss I made most of the changes you suggested. I found a better way to get the east and north vector components using 
`self.dataset_config.vector_variables[variable]['east_vector_component'], self.dataset_config.vector_variables[variable]['north_vector_component']` 
which should work for all magnitude variables. 

I also changed the code so it doesn't add the east and north vectors when saving a csv of the whole water column. We can reverse this in the future if requested by end-users pretty easily. 

Lastly, I don't think changing self.depths to self.depth on line 173 will affect a timeseries CSV over all depths since this line is only called when `self.depth != 'all'`. I didn't notice any issues when testing the code but if you know of a scenario where this might be a problem let me know and I'll re-evaluate.

JustinElms:
Tried re-basing the branch which made a mess of the commits. Will re-open as a new PR to keep history clean.

PR #847, opened 2021-05-27
Added interp property to datasetconfig to be able to set variable specific interpolation settings from datasetconfig

BaruaSourav:
## Background
The interpolation settings are selected on the settings tab of Navigator right now. When integrating the Potential Sub-surface channel (PSSC) variable for various datasets, it turned out that the PSSC variable needed to use the Nearest Neighbor method of interpolation for generating the tiles/creating the CSV column with rounded 0/1 values instead of float/fractional values. So, for doing that one have to change the interpolation method setting in the setting tab. 
Instead, we planned to be able to set the interpolation settings for specific variable from _datasetconfig.json_ file. This will give us the flexibility to set variable specific interpolation setting from the configuration file depending on how a user might want to visualize the variable on Navigator.

As we already have the _/variables_ endpoint return the variable specific properties, I added an _interpolation_ property to the _VariableConfig_ class and make that value returned when the Combobox first gets the list of variables after dataset change is detected. Then the React components were changed, so that, on a variable change it sets the interpolation method depending on what it has on _datasetconfig.json_ file.   

### What behavior of the application is changed?
After this change is merged and deployed, when a user selects a variable that has some interpolation settings set on the _datasetconfig.json_ file, the selected interpolation option will be changed according to the given interpolation settings for that variable. If there is no variable specific interpolation setting on the variable, the interpolation setting selected on the settings tab will be used. 
Presently, we only have variable specific interpolation setting for the _psubsurfacechannel_ variable of 1. GIOPS 10 Day Forecast dataset. And, to check if this works, user can select **Potential Sub-Surface Channel** variable for that dataset, and then going to the setting tab, user can see the interpolation method has been changed to **Nearest Neighbour** option. 
Also, if someone changes the variable again and that variable doesn't have any specific interpolation method mentioned in the config file, the interpolation method dropdown will reselect the previously selected value for interpolation method. 

## Checks
- [x] I ran unit tests.


oceannavigator/dataset_config.py
```diff
@@ -358,7 +358,15 @@ def is_zero_centered(self) -> bool:
             return from_config in ['true', 'True'] or from_config == True
         except KeyError:
             return False
-
+    @property
+    def interpolation(self) -> dict:
+        """
+        """
```
htmlboss, line 364: Don't forget the comment 😝

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -223,6 +232,7 @@ export default class OceanNavigator extends React.Component {
     }
     else {
       for (let i = 0; i < key.length; ++i) {
+        console.log(key[i]);
```
htmlboss, line 235:
```suggestion
```

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -166,14 +168,21 @@ export default class OceanNavigator extends React.Component {
...
+    else{
+      options = Object.assign({}, this.state.options);
+      this.prevOptions = this.state.options;
+      options.interpType = newOptions.interpType?newOptions.interpType:options.interpType;
+      options.interpRadius = newOptions.interpRadius?newOptions.interpRadius:options.interpRadius;
+      options.interpNeighbours = newOptions.interpNeighbours?newOptions.interpNeighbours:options.interpNeighbours;
+      options.mapBathymetryOpacity = newOptions.mapBathymetryOpacity?newOptions.mapBathymetryOpacity:options.mapBathymetryOpacity;
+      options.bathymetry = newOptions.bathymetry?newOptions.bathymetry:options.bathymetry;
+      options.topoShadedRelief = newOptions.topoShadedRelief?newOptions.topoShadedRelief:options.topoShadedRelief;
+      options.bathyContour = newOptions.bathyContour?newOptions.bathyContour:options.bathyContour;
```
htmlboss, line 184: Prefer to add whitespace around ternary operators for easier reading. Like so:

```suggestion
    if (newOptions == null) {
      options = this.prevOptions;
    }
    else {
      options = Object.assign({}, this.state.options);
      this.prevOptions = this.state.options;
      options.interpType = newOptions.interpType ? newOptions.interpType : options.interpType;
      options.interpRadius = newOptions.interpRadius ? newOptions.interpRadius : options.interpRadius;
      options.interpNeighbours = newOptions.interpNeighbours ? newOptions.interpNeighbours : options.interpNeighbours;
      options.mapBathymetryOpacity = newOptions.mapBathymetryOpacity ? newOptions.mapBathymetryOpacity : options.mapBathymetryOpacity;
      options.bathymetry = newOptions.bathymetry ? newOptions.bathymetry : options.bathymetry;
      options.topoShadedRelief = newOptions.topoShadedRelief ? newOptions.topoShadedRelief : options.topoShadedRelief;
      options.bathyContour = newOptions.bathyContour ? newOptions.bathyContour : options.bathyContour;
```

htmlboss (commented):
Left a couple of comments, but it's good you've managed to navigate the disaster of a front-end code base we have! TBH I consider the Javascript changes a hack until refactor it all. Ping me again for final approval when you're ready!

oceannavigator/dataset_config.py
BaruaSourav: Yeah. I missed it 😁 Added a comment.

BaruaSourav:
@htmlboss Yeah it is hard to navigate through the React component and follow the state propagation sometimes. The svg dependency [diagram](https://drive.google.com/file/d/1E1nNEbavcUL3_VViIOvQl-AHIWxR6rMg/view?usp=sharing) that @dwayne-hart generated by dependency cruiser is an awesome help for this. 
I have done the suggested changes, ready for a review.

htmlboss (approved):
Nice!

Issue #839, opened 2021-05-10
CSV file couldn't be downloaded when Nearest Neighbour  is selected as Colour Interpolation method


BaruaSourav:
**Describe the bug**
In the settings menu, the **Colour Interpolation** option determines how the color of the map will be interpolated. 

After selecting the "Nearest Neighbour" option in the interpolation method, the CSV file downloading option of Area plot is facing server error.  

**To Reproduce**
1. Go to the _Settings_ tab
2. Select Colour Interpolation method as **Nearest Neighbour**
![image](https://user-images.githubusercontent.com/12777798/117688215-24c2ff00-b193-11eb-9245-63470211a22d.png)

3. Go to global map and select an area using the _Area > Draw on map_ wait for the plot to show up

4. Below the plot. select _Save Image_ then _csv_

The API call faces **Internal Server Error** and Navigator will go to a white screen showing the error message. 

**Expected behavior**
When selecting the csv option, the csv file should be downloaded successfully. 

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 90.0.4430.72

**Additional context**
Below is the sentry issue that is logged when the server faces this issue. 
https://sentry.io/organizations/dfo-ocean-navigator/issues/2382877320/?project=5511557&referrer=slack

Issue #446, opened 2019-06-13
Add tests for config files


htmlboss:
In order to prevent missing variables or incorrect dataset settings in future deploys, we need to create new tests for datasetconfig.json and oceannanigator.cfg that Jenkins will run.

htmlboss:
Tests have been created.

PR #848, opened 2021-06-01
Modified the interpolation function to return a masked array when "Nearest Neighbour" is selected 

BaruaSourav:
## Background
Previously, when the user selected Nearest neighbor as interpolation method, then selected an area, the CSV file download feature was facing internal server error. The issue is addressed in [issue #839](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/839).

At _map.py_ file the below line #446 expects _self.data_ to be a _[MaskedArray]
(https://numpy.org/devdocs/reference/maskedarray.baseclass.html#numpy.ma.MaskedArray)_. 

https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/4beef8c5aef4c00a331c2c56d01503c1bf42777a/plotting/map.py#L446

But the [resample_nearest](https://pyresample.readthedocs.io/en/stable/api/pyresample.html#pyresample.kd_tree.resample_nearest) function of _pyresample_ module returns a _numpy.ndarray_ instead at - 

https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/4beef8c5aef4c00a331c2c56d01503c1bf42777a/data/netcdf_data.py#L606

So, I just converted the resampled _ndarray_ returned from pyresample.resample_nearest to a _np.ma.MaskedArray_ using [np.ma.asarray](https://numpy.org/doc/stable/reference/generated/numpy.ma.asarray.html#numpy.ma.asarray) while returning.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/bin/runserver.sh
+++ b/bin/runserver.sh
@@ -3,9 +3,9 @@
 # Usage: ./runserver.sh <port> <optional_dataset_config_file.json>
 # <optional_dataset_config_file.json>:  Specify a non-default dataset config file to load the Navigator with.
 #                                       Argument not required.
-WORKER_THREADS=5
+WORKER_THREADS=2
 
-[[ x"$1" == x"" ]] && PORT=5000 || PORT=$1
+[[ "$1" == "" ]] && PORT=5000 || PORT=$1
 
 if [ ! -e /usr/bin/nproc ] ; then
 
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -605,8 +605,8 @@ class NetCDFData(Data):
             # Nearest-neighbour interpolation (junk)
             elif self.interp == "nearest":
 
-                return pyresample.kd_tree.resample_nearest(input_def, data,
-                                                           output_def, radius_of_influence=float(self.radius), nprocs=8)
+                return np.ma.asarray(pyresample.kd_tree.resample_nearest(input_def, data,
+                                                           output_def, radius_of_influence=float(self.radius), nprocs=8))
 
         raise ValueError(f"Unknown interpolation method {self.interp}.")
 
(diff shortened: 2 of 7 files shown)


dwayne-hart:
Is there a way to change the nprocs=8 to a setting in the oceannavigator.cfg file? Or should a setting of 1 or 2 be used instead?

BaruaSourav:
@dwayne-hart The way it is done now, it doesn't get the value from the config file. But good point. @htmlboss, can we fetch a value from the oceannavigator.cfg file and use it as the value for _nprocs_ parameter here? I will take a look now.

JustinElms:
@dwayne-hart @BaruaSourav  Doug discusses this in his analysis doc and concluded that the value of `nprocs` has no effect on Navigator performance but suggests that we remove this argument to cleanup the code a bit.

BaruaSourav:
Okay, I think we should wait for @htmlboss to have a look at it.

htmlboss (approved):
Looks good! About the nprocs thing, I'd say it's safe to delete that. Dwayne's suggestion about making these things configurable is exactly what I've been thinking about for a while....maybe not for this use case since it seems to have no effect, but we should make better use of config files/environment variables over hard-coding things. It'll gave Dwayne a way easier time with fine tuning things depending on the hardware we're running on.

BaruaSourav:
I have removed the nprocs argument from the resampling functions that we are using from pyresample module. I also agree that using environment variables/config settings instead of using a hard coded value for places where it effects the performance, is a better idea to facilitate the independence of tuning this setting while deploying on different hardware.

dwayne-hart (approved):

PR #849, opened 2021-06-03
Add bearing to magnitude timeseries csv and ensure timeseries data is ordered 

JustinElms:
**Resubmission of PR#843 to avoid messy commit history

## Background
CCG requested that we include water velocity bearing in timeseries csv's. When working on adding this data to the csv output I also found two other problems that are addressed here:

1. The  `Save Image - CSV` function on the Virtual Mooring window gives an Internal Server Error in all instances.
2. The order of point data in the Virtual Mooring timeseries changed each time I restarted my local Navigator instance which lead to concerns about the integrity of VM data. This was also observed when comparing VM plots on the production and staging sites - recreating the same plot on each produced different results - and when changing the start and end times of the timeseries plots. 

It looks like work on including bearing in the water velocity csv output was started in the past so I built on the existing unfinished code pieces to complete the feature. In the `load_data` function  there was a check for 'mag' in the variable name to see if we were plotting water velocity. Here I added functionality to pull the east and north velocity component data using the 
 `get_timeseries_point` function then adding it as `self.quiver_data`. When the `csv` function is called it checks for this attribute and if present adds additional columns for the east and north vector components and bearing data to the csv. The code for the calculation work was already present so I leveraged that to create the data. The csv output does not include the east and north vectors when saving a csv of the whole water column. We can reverse this in the future if requested by end-users pretty easily. 

There other issues turned out to be quick fixes. 1 above was resolved by changing `self.depths` to `self.depth` on line 173. Issue 2 was caused using unsorted timestamps to get the nc file list then applying `list(set(self.__flatten_list(file_list)))` to remove duplicates from the list. Apparently this doesn't preserve the list order (TIL!). As of python 3.5 dictionaries maintain their order so this was changed to `list(dict.fromkeys(self.__flatten_list(file_list)))` which appears to produce nc file lists in a consistent order for the tests that I've run.

This gets the east and north vector components using 
`self.dataset_config.vector_variables[variable]['east_vector_component'], self.dataset_config.vector_variables[variable]['north_vector_component']` 
which should work for all magnitude variables since these attributes are given for all 'mag' varaibles. Once the additional data is collected I use the previously developed code to calculate the bearing for the csv output.  

I extracted timeseries data directly from the netCDF files to validate that the order is correct which it was for the datasets I tested against. 

## Checks
- [X] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/sqlite_database.py
+++ b/data/sqlite_database.py
@@ -63,13 +63,13 @@ class SQLiteDatabase:
             variable = ?
             AND timestamp = ?;
         """
-        for ts in timestamp:
+        for ts in sorted(timestamp):
             for v in variable:
                 self.c.execute(query, (v, ts))
                 file_list.append(self.__flatten_list(self.c.fetchall()))
 
-        # funky way to remove duplicates from the list: https://stackoverflow.com/a/7961390/2231969
-        return list(set(self.__flatten_list(file_list)))
+        # funky way to remove duplicates while preserving order: https://stackoverflow.com/a/39835527
+        return list(dict.fromkeys(self.__flatten_list(file_list))) 
 
     def get_all_dimensions(self) -> List[str]:
         """Returns a list of all the dimensions in the Dimensions table.
--- a/plotting/timeseries.py
+++ b/plotting/timeseries.py
@@ -105,26 +105,46 @@ class TimeseriesPlotter(PointPlotter):
             times = dataset.nc_data.timestamps[starttime_idx: endtime_idx + 1]
             if self.query.get('dataset_quantum') == 'month':
                 times = [datetime.date(x.year, x.month, 1) for x in times]
-
-            if 'mag' in variable:
+            
+            if 'mag' in variable and self.depth != 'all':
                 # Under the current API this indicates that velocity data is being
-                # loaded. Save each velocity component (X and Y) for possible CSV
-                # export later.
-                """
-                self.quiver_data = [point_data[:, 0, :], point_data[:, 1, :]]
+                # loaded. Save each velocity vectorcomponent (X and Y) for possible 
+                # CSV export later. Currently, we only provide velocity components 
+                # for a single depth. 
 
-                point_data = np.ma.expand_dims(
-                    np.sqrt(
-                        point_data[:, 0, :] ** 2 + point_data[:, 1, :] ** 2
-                    ), 1
-                )
-                """
+                vector_variables = [
+                    self.dataset_config.vector_variables[variable]['east_vector_component'],
+                    self.dataset_config.vector_variables[variable]['north_vector_component']
+                ]
+
+                self.vector_variable_names = self.get_variable_names(dataset, vector_variables)
+                self.vector_variable_units = self.get_variable_units(dataset, vector_variables)
+                                
+                d = []
+                vector_point_data = []
+                for vv in vector_variables:
+                    for p in self.points:
+                        vector_data = []
+                        
+                        d,_ = dataset.get_timeseries_point(
+                            float(p[0]),
+                            float(p[1]),
+                            self.depth,
+                            self.starttime,
+                            self.endtime,
+                            vv,
+                            return_depth=True
+                        )
+                    
+                    vector_data.append(d)
+                    vector_point_data.append(np.ma.array(vector_data))
+                
+                self.quiver_data = vector_point_data
 
             self.times = times
             self.data = point_data
             self.depths = depths
             self.depth_unit = "m"
-
 
     def csv(self):
         header = [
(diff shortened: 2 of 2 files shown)


data/sqlite_database.py
```diff
@@ -63,13 +63,13 @@ def get_netcdf_files(self, timestamp: List[int], variable: List[str]) -> List[st
             variable = ?
             AND timestamp = ?;
         """
-        for ts in timestamp:
+        for ts in sorted(timestamp):
```
htmlboss, line 66: Nice one.

htmlboss: Side note:
Now that I look at this function, the way it's requesting data from the DB is actually terrible because it's executing one query per timestamp per variable (yes I was the idiot who wrote this 2 years ago). SQL support multiple values in a single WHERE statement so it's really just a matter of passing `sorted(timestamp).join(',')` into `self.c.execute(...)` I think to eliminate the outer loop.

No need to action this since we're not experiencing performance issues yet, but this file is a prime candidate for optimization.

data/sqlite_database.py
```diff
@@ -63,13 +63,13 @@ def get_netcdf_files(self, timestamp: List[int], variable: List[str]) -> List[st
...
             for v in variable:
                 self.c.execute(query, (v, ts))
                 file_list.append(self.__flatten_list(self.c.fetchall()))
 
-        # funky way to remove duplicates from the list: https://stackoverflow.com/a/7961390/2231969
-        return list(set(self.__flatten_list(file_list)))
+        # funky way to remove duplicates while preserving order: https://stackoverflow.com/a/39835527
+        return list(dict.fromkeys(self.__flatten_list(file_list)))
```
htmlboss, line 72: Nit: Could we write a test to ensure the correct ordering for this in `tests/test_sqlite_database.py`? There might already be a test for that but I can't remember LOL

plotting/timeseries.py
```diff
@@ -105,27 +105,47 @@ def load_data(self):
...
+                self.vector_variable_names = self.get_variable_names(dataset, vector_variables)
+                self.vector_variable_units = self.get_variable_units(dataset, vector_variables)
+                                
+                d = []
+                vector_point_data = []
+                for vv in vector_variables:
+                    for p in self.points:
+                        vector_data = []
+                        
+                        d,_ = dataset.get_timeseries_point(
```
htmlboss, line 129:
```suggestion
                        d, _ = dataset.get_timeseries_point(
```

plotting/timeseries.py
```diff
@@ -137,34 +157,29 @@ def csv(self):
...
-            # velocity components (pulled from the quiver_data attribute) and
-            # bearing information (to be calculated below).
-            have_quiver = hasattr(self, 'quiver_data')
-        else:
-            have_quiver = False
+        
+        # Check to see if the quiver attribute is present. If so the CSV export will 
+        # also include X and Y velocity components (pulled from the quiver_data attribute) 
+        # and bearing information (to be calculated below).
+        have_quiver = hasattr(self, 'quiver_data')
```
htmlboss, line 164:
```suggestion
        has_quiver = hasattr(self, 'quiver_data')
```

plotting/timeseries.py
```diff
@@ -177,13 +192,15 @@ def csv(self):
...
+                    self.vector_variable_names[1], self.vector_variable_units[1],
                     "Bearing (degrees clockwise positive from North)"
                 )
                 header.append(["Variables", header_text])
-                for var_name in [self.variable_name, self.variable_names[0],
-                                 self.variable_names[1], "Bearing"]:
+                for var_name in [self.variable_name, 
+                                 self.vector_variable_names[0], 
+                                 self.vector_variable_names[1],
+                                 "Bearing"]:
```
htmlboss, line 203: Cool trick: you can use `*my_list` to unpack a list into another list :exploding_head: and write less code
```suggestion
                for var_name in [self.variable_name, *self.vector_variable_names, "Bearing"]:
```

Go to https://www.programiz.com/python-programming/online-compiler/
and paste and run to play with the syntax:
```python
test = [1, 2, 3]

more = [4, 5, *test]

print(more)

```

htmlboss (commented):
Glad to see this fixed! Left a few FYI's and one question about unit testing. Almost ready to approve.

JustinElms:
@htmlboss Thanks for the feedback! I made the revisions you suggested in timeseries.py. I'll work on a test to ensure the data pulled by get_netcdf_files is ordered correctly and look into reducing the number of queries that function is making by joining the timestamps.

htmlboss (approved):
Good to go!

PR #844, opened 2021-05-24
Add velocity arrows to openlayers

htmlboss:
## Checks
- [ ] I ran unit tests.
- [ ] I've tested the relevant changes from a user POV.

_Hint_ To run all python unit tests run the following command from the root repository directory:
```sh
bash run_python_tests.sh
```

PR #851, opened 2021-06-13
Implement bearing calculated variable

htmlboss:
## Background

Required for #844 

Needed this bearing calculation in another project so I figured I'd add it as a calculated variable so it can be displayed in the front-end for debugging and maybe some scientific purpose.

* Implements a bearing (degrees clockwise positive from North) calculated variable.

Copied the code in the CSV generation for bearing calculations.

Open to suggestions for colour map...https://matplotlib.org/cmocean/

## Screenshot(s)

image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/calculated_parser/functions.py
+++ b/data/calculated_parser/functions.py
@@ -52,9 +52,41 @@ def magnitude(a, b):
     b: ndarray
 
     Returns:
-        ndarray -- magnitude of a and b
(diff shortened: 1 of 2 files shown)


dwayne-hart (approved):
Clean implementation.

Issue #842, opened 2021-05-20
Requesting API script for pre-defined area plot gives 'Not Found' error


JustinElms:
**Describe the bug**
Users cannot download API scripts for area plots if they selected one of the pre-defined areas from the Area menu. Instead, they get:

Not Found 
The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.

This happens for many combinations of dataset , variable, and area. It appears that the API call is missing the `generatescript` endpoint due to a formatting issue when pre-defined areas are used. For areas that were drawn or enter the query looks as expected, like this:

api **/** v1.0 **/** generatescript **/** %7B%22area%22:[%7B%22innerrings%22:[],%22name%22:%22%22,%22polygons%22:[[[44.59046718130884,-53.296874463558204],[44.84029065139799,-45.21093696355822],[40.78054143186034,-45.12304633855821],[40.84706035607121,-52.9453119635582],[44.59046718130884,-53.296874463558204]]]%7D],%... **/** python **/** CSV **/**

but when a pre-defined are is used the area and it's set are written as NAFO_Divisions **/** 6H like so:

api **/** v1.0 **/** generatescript **/** %7B%22area%22:[%22NAFO_Divisions **/** 6H%22],%... **/** python **/** CSV **/**

the **/** character dividing the set and area appears to prevent the call from hitting the endpoint, replacing it with a **-** or similar character would likely fix the problem.    

**To Reproduce**
Select any of the pre-defined area sets from the drop-down menu the click an area on the global view. Once the area plot window opens and the data has loaded (click 'plot' on the Navigation bar if not) select any option from the 'API Script' drop-down list. 

**Expected Behaviour**
The user should receive a generated API script for the selected area. 

**Desktop:**
 - OS: Ubuntu 20.04 and Windows 10
 - Browser: Chrome
 - Version: 90.0.4430.212

PR #853, opened 2021-06-29
Refactoring the /generateScript API end-point to receive query params.

BaruaSourav:
## Background
When exploring the issue #842, I found that the API end point _/generateScript_ at  https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/897db66fade74abaa3ad6b504683302a3056b9dc/routes/api_v1_0.py#L87 
is receiving slash (/) separated values as parameters for the GET request. This caused an error in the API URL when it got generated for the predefined area, cause the predefined area names are formatted as  _NAFO_Divisions / 6H_. This '/' in the middle is causing the API URL mismatch and showing an error.

After discussion with @htmlboss, we decided to refactor the part of the API, so that it receives query parameters. I have changed the _/generateScript_ route on the API. Merging this PR will solve the issue of Pre-defined area not being able to generate API Script. 

While solving this issue I found out that we don't have a template for generating R API script for CSV file downloading. I will open an issue for this one and will add a template file in this location [/plotting/templates]( https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/tree/master/plotting/templates)

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


routes/api_v1_0.py
```diff
@@ -84,18 +84,29 @@ def test_sentry():
...
         b = generatePython(query, scriptType)
         resp = send_file(b, as_attachment=True,
-                         attachment_filename='script_template.py', mimetype='application/x-python')
+                         attachment_filename='API_script_'+scriptType+'.py', mimetype='application/x-python')
 
     elif lang == "r":
         b = generateR(query, scriptType)
         resp = send_file(b, as_attachment=True,
-                         attachment_filename='script_template.r', mimetype='application/x-python')
+                         attachment_filename='API_script_'+scriptType+'.r', mimetype='application/x-python')
```
htmlboss, line 109:
```suggestion
                         attachment_filename='API_script_'+scriptType+'.r', mimetype='text/plain')
```

htmlboss: R files don't appear to have their own MIME type so we can just use plain text.
 https://www.sitepoint.com/mime-types-complete-list/

oceannavigator/frontend/src/components/PlotImage.jsx
```diff
@@ -40,26 +40,16 @@ export default class PlotImage extends React.PureComponent {
...
-          url = window.location.origin + "/api/v1.0/generatescript/" + url +
-          `&save&format=csv&size=${this.props.query.size}` +
-          `&dpi=${this.props.query.dpi}` + "/r/" + "CSV/";
-        }
-      
+        var url = window.location.origin + "/api/v1.0/generatescript/?query=" +query+ "&lang=r" + "&scriptType=PLOT";
+      } else if (language == "pythonCSV") {
+        var url = window.location.origin + "/api/v1.0/generatescript/?query=" + query + "&lang=python" +"&scriptType=CSV";
+      } else if (language == "rCSV") {
+        var url = window.location.origin + "/api/v1.0/generatescript/?query=" + query+ "&lang=r" + "&scriptType=CSV";
```
htmlboss, line 51: Javascript support `switch` statements on strings so we could actually simplify this a bit further to reduce code duplication. For example:

```js
const query = encodeURIComponent(stringify(this.generateQuery(this.props.query)));

let scriptLang = null;
let scriptType = null;
switch (language) {
  case "pythonPlot" {
    scriptLang = "python";
    scriptType = "PLOT";
    break;
  }
  // Add other cases for rPlot, pythonCSV, and rCSV!


  default:
    console.error(`Unknown language {language}.`);
    break;
}

const url = `${window.location.origin}/api/v1.0/generatescript/?query=${query}&lang=${scriptLang}&scriptType=${scriptType}`;

```

Notice my use of back-ticks (`) to use string interpolation syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

htmlboss: As a side-note we avoid declaring new variables in Javascript using `var` since that's basically a global variable which is bad. https://chris-franklin.com/why-is-var-considered-a-bad-practice-in-javascript/

htmlboss (commented):
Great work with this! I just left a couple of minor comments along with few links for your learning. Tag me again for a final approval when you're ready.

routes/api_v1_0.py
BaruaSourav: Yes, right. Committing the change.

oceannavigator/frontend/src/components/PlotImage.jsx
BaruaSourav: Yes good one, switch case will be a perfect fit for this task. And, yes the variable needs to be declared as a _const_. Making the changes now.

BaruaSourav:
@htmlboss The changed has been committed, need your final feedback and then I believe this one is ready to be merged.  😄

htmlboss (approved):

PR #859, opened 2021-07-08
Exclude unknown fields in TimestampsSchema

htmlboss:
## Background

#857 exposed a frontend bug (surprise surprise) where jQuery is sending a hidden `_=1625711072192` parameter as part of the timestamps payload in `TimePicker.jsx`....not sure why it's doing that. Anyway there's no further security risk as we just silently ignore this unknown value. 

By default `marshmallow` raises an error when it encounters a URL parameter that hasn't been explicitly defined in the schema...a security feature. We've been accepting anything and everything up to this point so this bug has probably been around for several years...

* Followed: https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-unknown-fields

Tests would have never caught this anyway because this might be a jQuery bug...

## Screenshot(s)
Works now (note the `_` field on the right side of the URL):
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/routes/schemas/timestamps_schema.py
+++ b/routes/schemas/timestamps_schema.py
@@ -1,4 +1,5 @@
-from marshmallow import Schema, fields
+from marshmallow import EXCLUDE, Schema, fields
+
 
 class TimestampsSchema(Schema):
     """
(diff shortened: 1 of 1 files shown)


dwayne-hart (approved):

htmlboss:
I'm going to sleep now Dwayne so you can go sleep too :joy:

PR #863, opened 2021-07-14
Adding R script template for CSV download and verifying all the generated API script of Python and R are working

BaruaSourav:
## Background
There isn't any template for the R CSV API script, so trying to get an R API script for CSV from the plotting window would result into an error.

I am adding a template for that case following the existing template that we have for R-Plot and it is returning a script right now. 

However I have tried to run the returned R scripts and it turns out they aren't able to download the plot or csv as they intend to do and faces error. I am looking into editing the templates so that they work perfectly.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/yarn.lock
+++ b/oceannavigator/frontend/yarn.lock
@@ -1357,6 +1357,11 @@ camelcase@^2.0.0:
   resolved "https://registry.yarnpkg.com/camelcase/-/camelcase-2.1.1.tgz#7c1d16d679a1bbe59ca02cacecfb011e201f5a1f"
   integrity sha1-fB0W1nmhu+WcoCys7PsBHiAfWh8=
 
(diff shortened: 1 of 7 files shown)


htmlboss (commented):
Last thing is to delete this line:
https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/master/tests/routes/test_api_v_1_0_generatescript.py#L30

so now that test should all pass via `python -m unittest tests/routes/test_api_v_1_0_generatescript.py`

BaruaSourav:
I have removed the _unittest.skip_ line from that unit test file and ran the test. The test passed. 🎉 Btw, there are 3 commits for this single line deletion, as I tried to modify the commit message and turns out it just added a new commit!

htmlboss (approved):

PR #865, opened 2021-07-22
Update `webpack` to latest v4

htmlboss:
## Background

Followed webpack guides for v1->v2->v3->v4.

Latest v4 version is `4.46.0`

Avoiding v5 for now since some dependencies we use still aren't ready for v5. Plus v5 was only released during the last quarter of 2020 so it's still pretty new.

* _Dwayne I'd suggest doing a `yarn dev` and a `yarn build` in two separate fresh environments to ensure this crap is actually working before approving + merging._

* We might start getting more automatic dependabot updates now that we're out of ancient history for webpack...Plenty of other fossils to be updated though.

* `yarn build` WORKS AGAIN :partying_face: 

* Some of the dependencies have not been updated to their latest versions since they would depend on webpack 5. But they are at their latest compatible versions for webpack v4.

* Took a few hours but we have a net reduction in dependencies which is always good.

## Screenshot(s)

No difference to website. 

image

Successful `yarn build`
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/package.json
+++ b/oceannavigator/frontend/package.json
@@ -66,7 +66,6 @@
     "file-loader": "^0.9.0",
     "html-webpack-plugin": "^2.29.0",
     "i18next-conv": "^3.0.3",
-    "json-loader": "^0.5.4",
     "node-sass": "^4.13.1",
     "resolve-url-loader": "^1.6.0",
     "sass-loader": "^6.0.6",
--- a/oceannavigator/frontend/package.json
+++ b/oceannavigator/frontend/package.json
@@ -5,8 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "dev": "webpack -d --watch",
-    "build": "NODE_ENV=production node --stack_size=1000000 node_modules/.bin/webpack --progress -p",
+    "dev": "webpack --watch --mode=development",
+    "build": "NODE_ENV=production node --stack_size=1000000 node_modules/.bin/webpack --progress --mode=production",
     "i18n": "i18next-conv -s ../translations/en/LC_MESSAGES/messages.mo -l en -t ../translations/en.json && i18next-conv -s ../translations/fr/LC_MESSAGES/messages.mo -l fr -t ../translations/fr.json",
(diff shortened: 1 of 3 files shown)


oceannavigator/frontend/package.json
```diff
@@ -59,18 +59,17 @@
...
-    "extract-text-webpack-plugin": "^3.0.0",
-    "file-loader": "^0.9.0",
-    "html-webpack-plugin": "^2.29.0",
+    "css-loader": "^4.3.0",
+    "eslint": "^7.3.1",
+    "eslint-plugin-react": "^7.24.0",
+    "file-loader": "^6.2.0",
+    "html-webpack-plugin": "^4.5.2",
     "i18next-conv": "^3.0.3",
-    "json-loader": "^0.5.4",
```
htmlboss, line 69: `webpack` installs this as a dependency automatically now

oceannavigator/frontend/package.json
```diff
@@ -59,18 +59,17 @@
...
+    "mini-css-extract-plugin": "^1.6.2",
     "node-sass": "^4.13.1",
     "resolve-url-loader": "^1.6.0",
     "sass-loader": "^6.0.6",
-    "style-loader": "^0.13.1",
```
htmlboss, line 73: not used

oceannavigator/frontend/package.json
```diff
@@ -59,18 +59,17 @@
   },
   "devDependencies": {
     "babel-polyfill": "^6.26.0",
-    "css-loader": "^0.25.0",
-    "eslint": "^6.0.1",
-    "eslint-plugin-react": "^7.6.1",
-    "extract-text-webpack-plugin": "^3.0.0",
```
htmlboss, line 65: deprecated in favour of `mini-css-extract-plugin`

oceannavigator/frontend/yarn.lock
```diff
@@ -2,24 +2,24 @@
 # yarn lockfile v1
```
htmlboss, line 2: autogenerated file so this can be ignored.

dwayne-hart (approved):

PR #866, opened 2021-07-23
Update JS linting

htmlboss:
## Background

* Makes our code look consistent

* Leverage Google JS style guide

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


htmlboss:
Way out of date. Gonna close and restart with a slightly less destructive approach.

PR #867, opened 2021-07-23
Remove unused JS packages

htmlboss:
## Background

Delete stuff we're not using.
Not extensive but this is the low hanging fruit.

A bunch of these deps are already handled in the `yarn.lock` so we shouldn't be declaring them in `package.json`.

* Fewer things to download will result in a faster `yarn install` and `yarn build`.
* Reduces surface area for security vulnerabilities cuz JS is filled with them.
* Easier to keep packages up to date.

## Screenshot(s)

No difference to website user experience.
image
image
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/yarn.lock
```diff
@@ -1979,21 +1959,11 @@ css-select@^4.1.3:
     domutils "^2.6.0"
     nth-check "^2.0.0"
 
-css-what@^5.0.0, css-what@^5.0.1:
+css-what@^5.0.0:
```
htmlboss, line 1962: you can see here that `css-what` is already marked as a dependency by another package we use...no need to explicitly add it in `package.json`.

dwayne-hart (approved):

PR #873, opened 2021-08-03
Fix subsetting of predefined areas

JustinElms:
## Background
We found that users cannot save a NetCDF subset of a predefined area because the name of the area is passed to the `calculateAreaBoundingBox` method in AreaWindow.jsx which is expecting an object with polygons. This yields and error and prevents the Navigator from proceeding. When digging into this I also found that the _Save API Script_ functionality is also broken because it does not meet the schema of the `generate_script` end point. 

The first part of this issues was addressed by checking if the user was requesting a subset of predefined area. If so the query includes the area name instead of the polygon's bounds. On the backed we now check for the 'area' attribute in the query. If present we grab the area's polygon and calculate the bounds. The Navigator can then produce a NetCDF file for the requested area.

Similar steps were taken in generating the query for the API script and a 'SUBSET' parameter was added to the schema so that it accepts the query. The resulting API script produces a NetCDF file but both xarray and ncdump give 'unknown format' errors when trying to open the resulting product. This is not limited to API scripts generated from predefined areas. 

To generate a NC file we need the polygons of the predefined area. Rather than create a new API endpoint to produce this data it makes more sense to grab these on the back end when needed. 

The `subsetArea` and `saveScript` functions in AreaWindow.jsx use `window.location.href` to send the query to the API. I considered re-implementing this using jQuery but I decided to hold off for now since we're migrating to Axios. 

As I mentioned about, the product of the API scripts gives an error when opened with ncdrump or xarray. This needs to be addressed but rather than tack it onto this PR I'll create a separate issue. 

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


dwayne-hart:
@htmlboss what is the status of this review?

dwayne-hart (approved):

PR #883, opened 2021-08-06
[Major change] Dump `babel` for `typescript`

htmlboss:
## Background

* Deleted babel and all of it's dependencies.
* Implemented typescript code generation and sets our primary JavaScript language target to be ES6. We used to do ES5 with a boat load of slow polyfills.
* Updated i18n since it broke with the migration.

For the refactor, if we were to use the new features of JS (spread operator, async functions, etc) using Babel, I'd have to install a bajillion babel packages (one for each language feature), and set up too many flaky configuration files, so I said enough is enough and switched to typescript which handles code generation much more intelligently than babel. Plus, everyone on the planet is using TS now so it's a very strong community and very mature now compared to 3 years ago.

Figured that since our primary users are all now on Windows 10 (which comes with IE Edge, and Chrome by default), we now should leverage modern languages features to clean up our code and speed things up. IE 11 is history in terms of our user base (which is why we kept babel around for so long).

Followed: https://asabahebwa.medium.com/setting-up-typescript-for-a-modern-project-using-webpack-4-d628388e610c

I hate javascript
* Translations were deleted in some areas due to babel/webpack transpilation/injection issues and us not using `refs` correctly when initializing the main map...

* This marks the start of slow migration to typescript from jsx. 
* _I've set things up so there's no change to existing coding or building steps...it'll magically pick up existing *.jsx files and also *.ts(x) files as they're created_. So `yarn dev`, `yarn build`, `yarn install` all still work normally and *.jsx can still be written!
* Changes pose high risk for breaking user experience so extensive testing required before merge.
* NOTE THE AMOUNT OF DELETIONS COMPARED TO ADDITIONS IN THE TOP RIGHT WITH US DUMPING BABEL. ITS A NET REDUCTION EH.

## Screenshot(s)

Translations still work:
image
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/.babelrc
```diff
@@ -1,8 +1,14 @@
...
-  "env": {
-    "production": {
-      "presets" : ["react-optimize"]
-    }
-  }
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "entry"
```
htmlboss, line 6: https://github.com/zloirock/core-js#babelpreset-env

oceannavigator/frontend/package.json
```diff
@@ -62,6 +57,8 @@
     "mini-css-extract-plugin": "^1.6.2",
     "node-sass": "^4.13.1",
     "sass-loader": "^6.0.6",
+    "ts-loader": "^8.2.0",
```
htmlboss, line 60: Can't update to latest v9 cuz it doesn't support webpack 4 which is what we're currently on. no big deal tho https://github.com/TypeStrong/ts-loader/issues/595#issuecomment-827253722

oceannavigator/frontend/webpack.config.js
```diff
@@ -1,80 +1,80 @@
...
+  ],
+  output: {
+    path: BUILD_DIR,
+    filename: "oceannavigator.js",
+    publicPath: "/public/"
+  },
+  module: {
+    rules: [
+      {
+        test: /\.(t|j)sx?$/,
```
htmlboss, line 22: Progressive migration to typescript, so all of our existing `.jsx` files will still be automatically compiled.

oceannavigator/frontend/src/components/ImageSize.jsx
```diff
@@ -42,6 +42,7 @@ export default class ImageSize extends React.Component {
 
   render() {
     _("inches");
+    const _inches = _("inches");
```
htmlboss, line 45: todo: check if this is needed anymore since we're now on typescript.

jmunroe:
Thanks for taking the time to discuss this major change.  While I don't disagree with the motivation, I think this kind of substantial changes gives us a good opportunity to ensure we are making the best choices for the project. i.e. time well spent.

Can we separate the some of the  issues I am seeing being included in this PR?  
1. Which browser versions are we trying to support?
2. Updating the front end tool chain to use TypeScript instead of Babel for compiling our JSX.
3. Switch from JavaScript to TypeScript as the language for the front end development.

1.  Which browser versions are we trying to support?
I agree that we should be clear about the minimum requirement for a browser we need to support. This is of course strongly related which features of JavaScript ES6+/TC39 we are wanting to use in the project.  Through a combinations of polyfills and transpiling , tools like babel or typescript' compiler (among others) can be used allow the use of newer JavaScript features in older browser. 

2. Tool chain upgrades
JavaScript / front end development has seen an explosion of development and so there are a lot of solutions available to a lot of different problems.  When we look out changing the javascript transpiler from Babel to TypeScript and the associated packages, we are modifying that tool chain.  Since Ocean Navigator is primarily a ReactJS application, I think we should be sticking as close as possible to "official recommendations" for compiler / bundler / package manager rather than trying to roll our own.  The React developers have provided a tool [Create React App](https://reactjs.org/docs/create-a-new-react-app.html#create-react-app) that sets up a tool chain that is appropriate for a single page application.  

Have we tried running this create-react-app tool to set up all of the toolchain rather than trying to modify our existing toolchain?  I'll note that the "default" version of this tool still uses babel but it can also be used with typescript `npx create-react-app my-app --template typescript`.  

My understanding is Babel is still a widely used solution for front end development of ReactJS based single page application.  Babel has plugin-in system for different extensions that allows customization. I don't see continue to use Babel as inherently a "bad thing" assuming we are continuing to use packages with broad community support.  As such, the switch to TypeScript "just" to use the tsc compiler doesn't seem convincing.

3. Switch from JavaScript to TypeScript as the language for the front end development.
The biggest difference between JS and TS is clearly the introduction of a type system as opposed to being dynamically typed. This could result in the detection and removal of any number of unknown bugs in our frontend code. The developer tooling (using visual-studio-code) with TypeScript is supposedly really impressive as well.

Note, this is a different issue in needing to use TypeScript's tsc for the tool chain as opposed to using Babel.  I think even if we switched to TypeScript as a language we still have the choice of using Babel to transpile the TypeScript to JavaScript (using `@babel/preset-typescript` ).  

Up until recently, I think the Angular community and TypeScript have been very tightly coupled but I am also seeing more and more React/TypeScript related projects so perhaps there is no-issue at all to use TypeScript for the Navigator.

However, I don't this it is completely obvious that TypeScript is the only way to go for the project.  Here is an article on [7 Really Good Reasons Not to Use TypeScript (2020)](https://everyday.codes/javascript/7-really-good-reasons-not-to-use-typescript/).  One issue that jumps at mean is the "messiness" of the code.  My preference is code that is as clean / terse as possible. I like the readability and a real minimum of boilerplate.  (Yes, even if the cost is a dynamic type system!).  So, I'll ask: with TypeScript will code base be more readable and maintainable than keeping with JavaScript (ES6+) as our core language?  

=== 

In summary, I am not fundamentally opposed to switch to TypeScript over Babel / vanilla JS. But I do want to hash through possible implications and make sure there isn't even better option that is available to use at this time.  I would be concerned if our primary reason for going to TypeScript was "everyone on the planet is using TS".

Are there specific resources (books / websites / courses) that we can identify that explain clearly what the the state-of-the-art for a React based application using TypeScript?  For comparison, here are some React development books (JS based) that I've really liked:

- [FullStack React](https://www.newline.co/fullstack-react)
- [Road to React](https://www.roadtoreact.com/)

Is there are any up-to-date, recommend resources for TypeScript (and the associated tooling for developing single page applications) that others have used?   Just to answer my own question, a quick search revels that the following is now available and might be what I a thinking about:

- [FullStack React with TypeScript] [https://www.newline.co/fullstack-react-with-typescript)

htmlboss:
Been working off this typescript compiler branch for #880 including testing, and so far it's been a 100% drop-in replacement. No new bugs encountered. Frankly being able to target a specific JS version with a one-line config change has been a blessing.

I'm gonna make the call to merge this given timelines are closing to get the Go Button implemented.

dwayne-hart (approved):

PR #884, opened 2021-08-08
Re-enable main map quiver arrows

htmlboss:
## Background
Fixed the bug in the datasetconfig which caused the velocity arrows to point the wrong way. https://github.com/DFO-Ocean-Navigator/configurations/pull/27

Only way.

Did a bit of cleaning to hide the quiver selector in the point, line, and area windows since it's useless there.

## Screenshot(s)
Notice the arrows lining up with the area window plot. 

Main map (double resolution...every second point...this pr will deploy every fourth point):
image

Area window:
image

If you zoom out you can really see the current structure. really cool stuff
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


dwayne-hart (approved):

JustinElms:
This looks awesome!!

PR #889, opened 2021-08-11
Removing kml from git as it resides in [...]/data/kml on a datastore.

dwayne-hart:
Removing the linkage to GitHub's LSF service.


htmlboss (approved):
Noice just needs the config file change 🚀

PR #891, opened 2021-08-13
Add script to pre-gen tiles

htmlboss:
## Background
Applicable to all GIOPS and RIOPS datasets, at the surface, for the first timestamp found for each dataset (so for forecasts, today), for temp salinity water x water y and magnitude variables

Bit slow but I'm not sitting on top of the data. May take a couple of hours so best run during the night. We're paying for the electricity so might as well use it LOL.

We've got lots of RAM so let's warm up the cache to make stuff fast.

Invoked by `python scripts/pregen_tiles.py --url <url>`

* `python scripts/pregen_tiles.py --url http://10.5.166.251:5000`
* `python scripts/pregen_tiles.py --url https://navigator.oceansdata.ca`

Will be expanded in the future for other datasets, variables, depths, and times. Just getting some hack out quickly for a fast win.

Will execute remote requests in batches of whatever the number of vCPUs is available...8 on my system

## Screenshot(s)
n/a

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/scripts/pregen_tiles.py
+++ b/scripts/pregen_tiles.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+import argparse
+import multiprocessing
+from itertools import product
+
+import requests
+
+parser = argparse.ArgumentParser(description='Warm up the tile cache.')
+parser.add_argument('--url', type=str, required=True,
+                    help='Root URL to Navigator instance. Ex: https://navigator.oceansdata.ca or http://10.5.166.251:5000')
+args = parser.parse_args()
+
+base_url = f"{args.url}/api/v1.0"
+
+
+def get_tile(dataset: str, variable: str, time: int, depth: int, scale: str, x: int, y: int, z: int) -> None:
+    res = requests.get(
+        f"{base_url}/tiles/gaussian/25/10/EPSG:3857/{dataset}/{variable}/{time}/{depth}/{scale}/{z}/{x}/{y}.png", timeout=25)
+
+    if res.status_code != 200:
+        print(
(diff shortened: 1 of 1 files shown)


htmlboss:
Hope you enjoy Dwayne. Expect further improvements in the future.

scripts/pregen_tiles.py
```diff
@@ -0,0 +1,69 @@
...
+            f"Error getting tile {res.status_code} -- {res.url}")
+    else:
+        print(f"Fetched {res.url}")
+
+
+if __name__ == "__main__":
+    print(f"Starting with {multiprocessing.cpu_count()} processes...")
+
+    datasets = list(
+        filter(lambda d: "riops" in d or "giops" in d,
```
htmlboss, line 32: only do giops and riops for now since they're our highest traffic datasets.

scripts/pregen_tiles.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    timestamps = list(
+        map(lambda t: int(t['id']), requests.get(
+            f"{base_url}/timestamps/?dataset=giops_day&variable=votemper").json())
+    )
+
+    zoom_levels = [z for z in range(1, 8)]
+    x_range = [x for x in range(0, 8)]
+    y_range = [y for y in range(0, 8)]
```
htmlboss, line 45: this isn't the full openlayers grid but it covers most of what we want.

dwayne-hart (approved):

dwayne-hart:
I wish I had this before dory became unstable again :(

PR #895, opened 2021-08-31
Add Interactive Plotting to the Navigator

JustinElms:
## Background
Working towards bringing interactive plotting to the Navigator using the Plotly.js library.  This library helps to address some end-user feature requests and will allow them to  compare and manipulate plot data directly in their browsers.  So far, Plotly has been implemented in the Profile, CTD, Sound Speed, and Virtual Mooring plots of the point window. 

The Plotly and Bokeh libraries were considered for this feature. I tested Plotly.js as a front-end plotter and Bokeh at the back-end. Ultimately I opted to use Plotly for a number of reasons but mainly because having the plotter on the front-end allows us to push most of the plot generation to the client side which should improve performance.  

I setup a temporary API endpoint (plot_data) to provide data to the front-end plotter. My idea is to provide a stream of traces to the front-end so that the plot is populated quicker but performance is still slow. This is largely a result of having to open the dataset for every trace. 

The life cycle functions of the `PointPlotter` component could use an overhaul. The component works fine for a POC but I've noticed issues when changing variables in the Profile tab while data is being queried. 

## Screenshot(s)
Profile plotter with Temperature, Salinity, and sound speed data. The legend lists timestamps of available data from which users can slect one or more from to compare.
profile
Virtual Mooring/Timeseries plot of temperature data. Here the legend lists the available depths. 
TS

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/package.json
+++ b/oceannavigator/frontend/package.json
@@ -37,6 +37,7 @@
     "jquery-ui-month-picker": "kidsysco/jquery-ui-month-picker",
     "ol": "^6.6.1",
     "papaparse": "^5.2.0",
+    "plotly.js": "^2.3.1",
     "proj4": "^2.7.5",
     "prop-types": "^15.5.10",
     "rc-slider": "^9.2.2",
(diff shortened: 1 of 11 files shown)


JustinElms:
Closed until users request this feature

PR #911, opened 2021-09-23
Stop duplicate API calls for Variables dropdown

htmlboss:
## Background

* Re-jigged the state update logic in `OceanNavigator` to be more understandable.
* Moved the update interpolation options code to the `OceanNavigator` component to keep our `SelectBox` component as generic as possible.
* Added a new field (`two_dimensional`) to the `variables` API response to allow the front end to filter 3D or 2D variables without requiring another API call.
* Small refactor of the variables API endpoint to simplify conditional logic and remove code duplication.
* Fixed a bug in the `Map` component where the colourmap bar wouldn't be created when the website first loads. This bug was introduced when I simplified the state update logic in `OceanNavigator` to follow React best practices, so technically the fact the colourmap bar was showing before my refactor is actually a bug.
* Untangled some crap with the Range` component being tied to multiple variable scales in the React state which broke the variable dropdown for some reason.

Similar approach as #880 

nope

## Screenshot(s)
image

image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map.jsx
@@ -170,6 +170,10 @@ export default class Map extends React.PureComponent {
       location: [0, 90],
     };
 
+    this.scaleViewer = new app.ScaleViewer({
+      image: "/api/v1.0/scale/giops_day/votemper/-5,30.png",
+    });
+
     this.loader = function (extent, resolution, projection) {
       if (this.props.state.vectortype) {
         let url = "";
@@ -671,6 +675,7 @@ export default class Map extends React.PureComponent {
         }),
       ])
     });
+    this.map.addControl(this.scaleViewer);
     this.map.on("moveend", this.refreshFeatures.bind(this));
     this.map.on("moveend", function () {
       const c = olproj.transform(this.mapView.getCenter(), this.props.state.projection, "EPSG:4326").map(function (c) { return c.toFixed(4); });
@@ -1318,8 +1323,8 @@ export default class Map extends React.PureComponent {
   }
 
   shouldUpdateScaleViewer(currentDataset, prevDataset,
-                          currentVariable, prevVariable,
-                           currentScale, prevScale)
+    currentVariable, prevVariable,
+    currentScale, prevScale)
   {
     return (currentDataset !== prevDataset)
           || (currentVariable !== prevVariable)
@@ -1330,11 +1335,17 @@ export default class Map extends React.PureComponent {
     if (this.scaleViewer != null) {
       this.map.removeControl(this.scaleViewer);
     }
+
+    let scale = currentScale;
+    if (Array.isArray(scale)) {
+      scale = scale.join(",");
+    }
+
     this.scaleViewer = new app.ScaleViewer({
       image: (
         `/api/v1.0/scale/${currentDataset}` +
         `/${currentVariable}` +
-        `/${currentScale}.png`
+        `/${scale}.png`
       )
     });
     this.map.addControl(this.scaleViewer);
@@ -1404,6 +1415,12 @@ export default class Map extends React.PureComponent {
     const datalayer = this.map.getLayers().getArray()[1];
     const old = datalayer.getSource();
     const props = old.getProperties();
+
+    let scale = this.props.scale;
+    if (Array.isArray(scale)) {
+      scale = scale.join(",");
+    }
+
     props.url = "/api/v1.0/tiles" +
       `/${this.props.options.interpType}` +
       `/${this.props.options.interpRadius}` +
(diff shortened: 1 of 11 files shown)


oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -170,6 +170,10 @@ export default class Map extends React.PureComponent {
       location: [0, 90],
     };
 
+    this.scaleViewer = new app.ScaleViewer({
+      image: "/api/v1.0/scale/giops_day/votemper/-5,30.png",
+    });
```
htmlboss, line 175: Create a default colourmap bar when page first loads using default dataset and variable.

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -1330,11 +1335,17 @@ export default class Map extends React.PureComponent {
     if (this.scaleViewer != null) {
       this.map.removeControl(this.scaleViewer);
     }
+
+    let scale = currentScale;
+    if (Array.isArray(scale)) {
+      scale = scale.join(",");
+    }
```
htmlboss, line 1342: workaround for when the variable colour scale can be a string or an array...it should be an array at all times but fixing that entirely is for another time.

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -44,9 +44,13 @@ class OceanNavigator extends React.Component {
 
     ReactGA.ga("send", "pageview");
 
+    this.DEF_INTERP_TYPE = Object.freeze("gaussian");
+    this.DEF_INTERP_RADIUS_KM = Object.freeze(25);
+    this.DEF_INTERP_NUM_NEIGHBOURS = Object.freeze(10);
```
htmlboss, line 49: interpolation defaults

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -209,21 +210,56 @@ class OceanNavigator extends React.Component {
...
+          if (value.variable !== this.state.variable) {
+            const variableFromCache = this.state.datasetVariablesCache[value.dataset]
+            .find(v => v.id === value.variable);
+
+            // Update interpolation
+            newState["dataset_0"]["options"] = {
+              ...this.state.options,
+              interpType: variableFromCache.interp?.interpType || this.DEF_INTERP_TYPE,
+              interpRadius: variableFromCache.interp?.interpRadius || this.DEF_INTERP_RADIUS_KM,
+              interpNeighbours: variableFromCache.interp?.interpNeighbours || this.DEF_INTERP_NUM_NEIGHBOURS,
```
htmlboss, line 256: `?.`  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

htmlboss: This syntax I'm using here is short-circuit assignment: `expression || expression` will use the first `expression` that returns a truthy value:

`null || this.DEF_INTERP_TYPE` will return `this.DEF_INTERP_TYPE`.

`"gaussian" || this.DEF_INTERP_TYPE` will return "gaussian".

https://dev.to/flexdinesh/short-circuit-assignment-in-javascript--4k80

oceannavigator/frontend/src/components/Range.jsx
```diff
@@ -18,7 +18,7 @@ class Range extends React.Component {
 
     // Parse scale tuple
     let scale = this.props.state;
-    if (typeof (this.props.state.split) === "function") {
+    if (typeof this.props.state === "string" || this.props.state instanceof String) {
```
htmlboss, line 21: This is the correct way of checking if a variable is a string.

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -209,21 +210,56 @@ class OceanNavigator extends React.Component {
...
             this.changeDataset(value.dataset, value);
             return;
           }
+          
+          // If variable is being changed...
+          if (value.variable !== this.state.variable) {
+            const variableFromCache = this.state.datasetVariablesCache[value.dataset]
+            .find(v => v.id === value.variable);
+
+            // Update interpolation
```
htmlboss, line 251: cc @JustinElms I moved your fix into here

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -209,21 +210,56 @@ class OceanNavigator extends React.Component {
...
+            newState["dataset_0"]["options"] = {
+              ...this.state.options,
```
htmlboss, line 253: `...` spread syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

oceannavigator/frontend/src/components/OceanNavigator.jsx
htmlboss: we use this conditional because in the API, variables don't define interp options by default (i.e. `interp === null`) since they're all using the same options by default.

JustinElms (approved):
Nice work! Much easier to understand the logic here and I really like the addition of the `two_dimensional` variables response.

PR #916, opened 2021-10-17
Eliminate duplicate API calls in `DatasetSelector` component

htmlboss:
## Background

This finishes the effort to reduce the number duplicate API calls made to the back end in the context of preparing for the go button. There is still considerable work left to completely eliminate this problem, but there is indeed a noticeable performance improvement with the caching implementation and improved general reliability.

* Updated the UX of the loading pop-up window to be more descriptive (see below for screenshots). In the future, we could also display errors relating to the dataset change operation.
* Implemented the `axios-extensions` package to provide caching middle ware for some of our requests.
* Removed some implementations of the deprecated React function `componentWillReceiveProps` to no effect to functionality. There are still some areas where we use it, and they will need to be removed before React can be updated.
* Created a "defaults" file for where various components can reference the same default values (`oceannavigator/frontend/src/components/Defaults.js`) without copy-pasting stuff.
* Cleaned up the `PropTypes` in relevant files to improve the self-documentation ability of the code, and provide better early warnings to required props. Still lots of work to be done here in the future.
* Fixed a bug in `oceannavigator/frontend/src/remote/OceanNavigator.js` where `axios` was always redirecting via 301 due to a missing `/` in the URL, which resulted in the cache not working correctly. Now it works just fine.
* Added a new field to the `variables` API end point that indicates whether a variable is 2D or not. Used by the front end to filter available variables for various plots.
* Refactored dataset fetching (the old `change_dataset` function) from the `OceanNavigator` component into the `DatasetSelector` component to simplify the state update process amongst all of our components to make way for the go button. This is the bulk of the logic change. Most of the other changes in here are to make this refactoring work (lmao). Also, we now enforce a strict ordering by which we fetch dataset information. This will enable the `PlotImage` component to be improved in the future so that it doesn't spam plot requests with garbage data. If an error is encountered at any point of the loading stage, the process will abort and not change the dataset (9 times out of 10 anyway lol). This is to help reduce the chances of the UI getting in a bad state and then sending requests with garbage API arguments.
  1. Get `/datasets`
  2. Get `/variables` for selected dataset.
  3. Get `/depths` for selected dataset
  4. (Still in the stupid `Timepicker` component)...get `/timestamps` for selected dataset. We still need to refactor the `Timepicker` component and move all of it's timestamp loading logic (after simplifying it) into the `DatasetSelector` component.

* Used newer React and JS language features to make the code more idiomatic and less messy:
  * Replaced some `function` definitions and `.bind(this)` calls with arrow functions, which provide more compact syntax, and automatically bind the parent object.
  * Used call-backs with `setState` to ensure correct ordering of state change operations and avoid race conditions.
  * Heavy use of the spread operator `...`.

* I basically ripped out and re-wrote the core data loading logic in the front end. There are definitely hidden bugs in here that will take time to be weeded out. I tested the profiles, transect, hovmoller, and area plots with a few datasets to ensure nothing was blatantly busted however.
* The meat-and-potatoes of this change is in the `DatasetSelector` component (`oceannavigator/frontend/src/components/DatasetSelector.jsx`).
* Inspecting the changes commit-by-commit would be an okay way of breaking this pr into smaller chunks. I tried to keep the commits self-contained but this is the Navigator so....(start at 0b05595b7e8fe08276c29fe57b8bba10fd108dee if you want to do this).

## Screenshot(s)

Note the difference in speed at which the datasets are changed, and the difference between the number of requests sent just for the dataset variable. Not sure why the videos are sped up but whatever.

Current experience on production:

https://user-images.githubusercontent.com/5572045/143961221-ed1a5709-97bc-483f-be49-bac22f01cefc.mp4

image

After this change:

https://user-images.githubusercontent.com/5572045/143961456-8fcd7314-4aee-4ac6-96b7-f82ff107955e.mp4

image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/DatasetSelector.jsx
+++ b/oceannavigator/frontend/src/components/DatasetSelector.jsx
@@ -1,55 +1,257 @@
-/* eslint react/no-deprecated: 0 */
-
 import React from "react";
-import ComboBox from "./ComboBox.jsx";
-import TimePicker from "./TimePicker.jsx";
 import PropTypes from "prop-types";
-import VelocitySelector from "./VelocitySelector.jsx";
+import { Modal, ProgressBar } from "react-bootstrap";
+
+import TimePicker from "./TimePicker.jsx";
+import Range from "./Range.jsx";
 import SelectBox from "./lib/SelectBox.jsx";
+
+import {
+  GetDatasetsPromise,
+  GetVariablesPromise,
+  GetTimestampsPromise,
+  GetDepthsPromise
+} from "../remote/OceanNavigator.js";
+
+import { DATASET_DEFAULTS } from "./Defaults.js";
 
 import { withTranslation } from "react-i18next";
 
-// Default properties for a dataset-state
-const DATA_ELEMS = [
+const PARENT_ATTRIBUTES_TO_UPDATE = Object.freeze([
   "dataset",
-  "dataset_attribution",
-  "dataset_quantum",
+  "quantum",
   "variable",
   "variable_scale", // Default range values for variable
   "depth",
   "time",
   "starttime",
   "quiverVariable",
-];
+]);
 
 class DatasetSelector extends React.Component {
   constructor(props) {
     super(props);
 
+    // TODO: Move these into Default.js
+    this.DEF_INTERP_TYPE = Object.freeze("gaussian");
+    this.DEF_INTERP_RADIUS_KM = Object.freeze(25);
+    this.DEF_INTERP_NUM_NEIGHBOURS = Object.freeze(10);
+
+    this.state = {
+      loading: false,
+      loadingPercent: 0,
+      loadingTitle: "",
+      datasetVariables: [],
+      datasetTimestamps: [],
+      datasetDepths: [],
+      options: {
+        ...props.options,
+      },
+      ...DATASET_DEFAULTS,
+    };
+
     // Function bindings
-    this.variableUpdate = this.variableUpdate.bind(this);
     this.onUpdate = this.onUpdate.bind(this);
   }
 
-  variableUpdate(key, value) {
-    this.props.onUpdate("setDefaultScale", true);
-    this.onUpdate(key, value);
+  changeDataset(newDataset, currentVariable) {
+    const currentDataset = this.state.availableDatasets.filter((d) => {
+      return d.id === newDataset;
+    })[0];
+
+    this.setState({
+      loading: true,
+      loadingPercent: 10,
+      loadingTitle: `${currentDataset.value}`
+    });
+
+    const quantum = currentDataset.quantum;
+
+    GetVariablesPromise(newDataset).then(variableResult => {
+      this.setState({ loadingPercent: 33 });
+
+      // Carry the currently selected variable to the new
+      // dataset if said variable exists in the new dataset.
+      let newVariable = currentVariable;
+      let newVariableScale = this.state.variable_scale;
+      let interpType = this.state.options.interpType;
+      let interpRadius = this.state.options.interpRadius;
+      let interpNeighbours = this.state.options.interpNeighbours;
+      const variableIds = variableResult.data.map(v => { return v.id; });
+      if (!variableIds.includes(currentVariable)) {
+        newVariable = variableResult.data[0].id;
+        newVariableScale = variableResult.data[0].scale;
+        interpType = variableResult.data[0].interp?.interpType || this.DEF_INTERP_TYPE;
+        interpRadius = variableResult.data[0].interp?.interpRadius || this.DEF_INTERP_RADIUS_KM;
+        interpNeighbours = variableResult.data[0].interp?.interpNeighbours || this.DEF_INTERP_NUM_NEIGHBOURS;
+      }
+
+      // eslint-disable-next-line max-len
+      GetTimestampsPromise(newDataset, newVariable).then(timeResult => {
+        this.setState({ loadingPercent: 75 });
+        
+        const timeData = timeResult.data;
+
+        const newTime = timeData[timeData.length - 1].id;
+        const newStarttime = timeData[0].id;
+
+        // eslint-disable-next-line max-len
+        GetDepthsPromise(newDataset, newVariable).then(depthResult => {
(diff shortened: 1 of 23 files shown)


oceannavigator/frontend/src/components/LineWindow.jsx
```diff
@@ -65,29 +64,6 @@ class LineWindow extends React.Component {
     this._mounted = false;
   }
 
-  componentWillReceiveProps(props) {
```
htmlboss, line 68: `componentWillReceiveProps` is deprecated in newer React versions so this needs to be removed. Oddly enough deleting this had no noticeable difference on functionality :shrug:

oceannavigator/frontend/src/components/LineWindow.jsx
```diff
@@ -312,29 +288,17 @@ class LineWindow extends React.Component {
       <Panel.Collapse>
         <Panel.Body>
           <DatasetSelector
-            key='dataset_0'
+            key='line_window_dataset_0'
```
htmlboss, line 291: This `key` must be unique between all rendered `DatasetSelector`s so it's state gets updated correctly.

oceannavigator/frontend/src/components/MapInputs.jsx
```diff
@@ -268,27 +218,17 @@ class MapInputs extends React.Component {
 
 //***********************************************************************
 MapInputs.propTypes = {
-  state: PropTypes.object,
```
htmlboss, line 271: Remove a bunch of unused crap from the props and proptypes across several files.

oceannavigator/frontend/src/components/TimePicker.jsx
```diff
@@ -117,142 +109,134 @@ class TimePicker extends React.Component {
...
-                  );
-                  break;
-              }
-              this.pickerChange();
-            }.bind(this)
+    // eslint-disable-next-line max-len
+    GetTimestampsPromise(props.dataset, props.variable).then(timestampResult => {
+      const data = timestampResult.data;
+
+      let map = {};
```
htmlboss, line 116: all below this was copy-pasted from the original implementation, I just wrapped it all in the cache-able timestamps promise.

oceannavigator/frontend/src/components/TimePicker.jsx
```diff
@@ -117,142 +109,134 @@ class TimePicker extends React.Component {
...
+
+      this.setState(
+        {
+          data: data,
+          map: map,
+          revmap: revmap,
+          min: min,
+          max: max
+        },
+        () => {
```
htmlboss, line 150: Switch to using arrow function so we don't need to explicitly call `.bind(this)` since arrow functions are automatically bound to the parent object

oceannavigator/frontend/src/components/lib/SelectBox.jsx
```diff
@@ -26,17 +26,24 @@ export default class SelectBox extends React.Component {
...
       <FormGroup controlid={`formgroup-${this.props.id}-selectbox`}>
         <ControlLabel>{this.props.label}</ControlLabel>
         <FormControl
           componentClass="select"
           name={this.props.name}
           placeholder={disabled ? _("Loading...") : this.props.placeholder}
-          onChange={(e) => this.props.onChange(e.target.name, e.target.value)}
+          onChange={(e) => {
+            if (this.props.multiple) {
+              this.props.onChange(e.target.name, e.target.selectedOptions);
```
htmlboss, line 38: multiple selected options are stored in a different variable than one.

oceannavigator/frontend/src/remote/OceanNavigator.js
```diff
@@ -2,21 +2,23 @@
...
 const axios = require("axios");
+import { cacheAdapterEnhancer } from "axios-extensions";
 
-function _createPromise() {
-  return axios.create();
-}
+const instance = axios.create({
+  headers: { "Cache-Control": "no-cache" },
+  adapter: cacheAdapterEnhancer(axios.defaults.adapter)
+});
```
htmlboss, line 10: switch to one instance so the cache is shared amongst all promises.

oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -382,12 +382,17 @@ class DatasetSelector extends React.Component {
         options = this.state.datasetVariables;
       }
 
+      // Work-around for when someone selected a plot that requires
+      // 3D variables, but the selected dataset doesn't have any LOL.
+      // This check prevents a white-screen crash.
+      const stillHasVariablesToShow = options.length > 0;
```
htmlboss, line 388: Instead of a white-screen crash, we'll force a server-side error so we get a more contextually useful message like so 
image

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -671,6 +675,7 @@ export default class Map extends React.PureComponent {
         }),
       ])
     });
+    this.map.addControl(this.scaleViewer);
```
htmlboss, line 678: This should always be explicitly added to the map.

routes/api_v1_0.py
```diff
@@ -207,27 +206,24 @@ def variables_query_v1_0():
...
             data.append({
-                'id': variable,
-                'value': config.variable[variable].name,
-                'scale': config.variable[variable].scale,
-                'interp':config.variable[variable].interpolation
+                'id': v.key,
+                'value': config.variable[v].name,
+                'scale': config.variable[v].scale,
+                'interp':config.variable[v].interpolation,
+                'two_dimensional': v.is_surface_only()
```
htmlboss, line 225: API improvement: add extra information to API call so front-end can use it for filtering

htmlboss:
I see all of the conflicts. I'll address any feedback first, squash all of my commits down to 1, then rebase and resolve the conflicts. Otherwise, I'd spend the next month working through the conflict editor.

htmlboss:
rebase went completely bonkers. gonna put up another fresh pr shortly...

htmlboss:
#926 replaces this one without the busted rebase

Issue #897, opened 2021-09-01
Cannot subset NetCDF for an area with multiple variables


JustinElms:
**Describe the bug**
Users get an 'Internal Service Error' when sub-setting  a NetCDF file with multiple variables. This occurs for any combination of variables,  all types of areas (drawn, entered, uploaded, and predefined), and all of the datasets I tested with (GIOPS 1, RIOPS 7,  CCG 5, CIOPS 9, and  CMEMS 16).  The 'Compress as *.zip' option has no effect on outcome. 

**To Reproduce**
Select any dataset then produce an area plot using any method. Once the plot has loaded select multiple variables in the Subset panel and click save. This results in an 'Internal Server Error' page. 

**Expected behavior**
Either a NetCDF file or a zip file containing one should be downloaded after completing the above steps. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04 / Windows 10
 - Browser: Chrome / Firefox / Edge
 - Version: 92.0.4515.159 / 91.0.2 / 90.0.818.41

JustinElms:
Some of this behaviour appears to be related to Issue #919, while other instances are due to conflicting unnamed dimensions which get labelled `'dims_0', ''dims_1', 'dims_2', ... `.

PR #920, opened 2021-11-05
Restore NC3 Subsetting for Calculated Variables

### LINKED ISSUE #919: Cannot subset areas in NC-3 or NC-4 Classic formats.
**Describe the bug**

The _Subset NetCDF_ function seems to fail for some variables when subsetting to NetCDF-3 Classic, NetCDF-3 64-bit, and NetCDF-4 Classic formats resulting in an `Internal Server Error` or `502 Bad Gateway` error message. The default NetCDF-4 and NetCDF-3 NC formats appear to be unaffected.  This error occurs for at least the Temperature and Water Velocity variables in datasets 1, 5, and 6. Location and method use to select the area and compressing as Zip do not appear to affect this behaviour. I get the following traceback when reproducing in a development environment:

```
  File "<string>", line 2, in <module>
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/core/dataset.py", line 1232, in to_netcdf
    compute=compute)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/backends/api.py", line 747, in to_netcdf
    unlimited_dims=unlimited_dims)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/backends/api.py", line 790, in dump_to_store
    unlimited_dims=unlimited_dims)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/backends/common.py", line 266, in store
    unlimited_dims=unlimited_dims)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/backends/common.py", line 304, in set_variables
    name, v, check, unlimited_dims=unlimited_dims)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/backends/netCDF4_.py", line 473, in prepare_variable
    _set_nc_attribute(nc4_var, k, v)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.6/site-packages/xarray/backends/netCDF4_.py", line 287, in _set_nc_attribute
    obj.setncattr_string(key, value)
  File "netCDF4/_netCDF4.pyx", line 4088, in netCDF4._netCDF4.Variable.setncattr_string
OSError: file format does not support NC_STRING attributes
```

When comparing the subset xarray objects produced from the temperature, salinity, and water velocity data I noticed that the temperature and water velocity DataArrays have a 'dims' attribute with value `['time', 'depth', 'latitude', 'longitude']` that salinity does not have. Once this attribute is deleted from the affected variables, xarray is able to write the object to either NetCDF format. It seems that xarray isn't quite sure how to handle lists of strings such as this one, and tries to encode them as NC_STRINGs regardless of the desired output format. NC_STRING is a new data type only compatible with NetCDF-4 format which is why we're getting this error.  

Some discussion on how xarray handles and encodes string data can be found here: https://github.com/pydata/xarray/issues/2059

**To Reproduce**
Select an area on the main map using any method. Once the Area Window appears and the _Subset_ variable list has been populated create a subset of either Temperature or Water Velocity with NC3 Classic, NC3 64 bit, or NC4 Classic output formats. 

**Expected behavior**
The resulting NC file should download.  

**Desktop (please complete the following information):**
 - OS: [Ubuntu/Windows 10]
 - Browser [Chrome]
 - Version [95.0.4638.54/94.0.4606.61]

**Additional context**
I'm sure other datasets/variables are affected by this issue but haven't had the chance to verify yet. I'll update this issue as necessary.

JustinElms:
## Background
We were notified that users are unable to subset area data using NetCDF-3 formats for calculated variables. When creating a subset from one of these variables the Navigator attempts to add the attributes from the datasetconfig.json as variable attributes in the resulting NetCDF file. This includes the 'dims' attribute is a list of strings describing the dimensions used for the variable calculation. It seems that xarray will try to encode variables like this one using the _NC_STRING_ format which is only compatible with NetCDF-4 files, causing the error. More detail can be found in Issue #919.

Closes #919 

My initial approach was to cast the 'dims; variable to another datatype to see if xarray would encode it using a compatible format however nothing seemed to work. In the end I felt that it was best to drop this attribute entirely since the dimensions should be listed in the resulting NC file anyway. All calculated variables given in the datasetconfig.json file need to have a 'dims' attribute so we shouldn't need to add logic to check for this attribute before dropping it. 

I found that some of the calculated variables are returned without named dimensions resulting in variables with dimensions `dim_0, dim_1, etc.` so having a 'dims' attribute could help here. Since this seems to be the root of issue #897 I decided to address that in the fix I'm working on there.  

## Checks
- [X] I ran unit tests.

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -371,8 +371,8 @@ class NetCDFData(Data):
                                               y_coord: y_slice,
                                               x_coord: x_slice
                                           })})
-                # Remove 'dims' attr (prevents exporting to NC3 formats)            
-                subset[variable].attrs.pop('dims')  
+                # Convert 'dims' attr to string (allows exporting to NC3 formats)            
+                subset[variable].attrs['dims'] = "(" + ",".join(subset[variable].attrs['dims']) + ")"
 
         output_format = query.get('output_format')
         filename = dataset_name + "_" + "%dN%dW-%dN%dW" % (p0.latitude, p0.longitude, p1.latitude, p1.longitude) \
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -371,6 +371,8 @@ class NetCDFData(Data):
                                               y_coord: y_slice,
                                               x_coord: x_slice
                                           })})
+                # Remove 'dims' attr (prevents exporting to NC3 formats)            
+                subset[variable].attrs.pop('dims')  
 
         output_format = query.get('output_format')
         filename = dataset_name + "_" + "%dN%dW-%dN%dW" % (p0.latitude, p0.longitude, p1.latitude, p1.longitude) \


nsoontie (commented):
This looks like a reasonable solution. Could you clarify if the issue is for the "NetCDF-3 Classic" option or the "NetCDF-3 NC" option.

Before we merge this, should we check with end users to confirm that removing the dims attribute is acceptable? Perhaps by sharing a sample file?

JustinElms:
@nsoontie I've noticed this problem when using NetCDF-3 Classic, NetCDF-3 64-bit, and NetCDF-4 Classic but not the default NetCDF-4 or NetCDF-3 NC formats. It looks like we essentially build the variables manually without the `'dims'` attribute for NetCDF-3 NC and NetCDF-4 supports NC_STRINGs so it's not be a problem there.

We can certainly check with the end users to see if the `'dims'` attribute is necessary. My thinking was that a well constructed NetCDF file should have the dimensions defined so including them as an attribute was a little redundant but we can probably dump the dimension names into a single string using something like `join`. This should allow us to keep the `'dims'` attribute without raising the NC_STRING error.

nsoontie (approved):

PR #924, opened 2021-11-19
Fix subsetting for multiple variables and over time ranges

JustinElms:
## Background
Continuing the work started in PR #920 which allowed users to export subsets in all NetCDF formats but didn't fix issues that prevented users from creating subsets with multiple variables or over a time range. 

A major component of this issue was the `calculated.isel()` function which returned xarray DataArray objects without properly defined dimensions. These objects were given xarray's default dimension names (`[dim_0, dim_1,...]`) which caused conflicts when variables of varying sizes were added to the subset xarray object. Generally this occurs when sub-setting 2D and 3D calculated variables. In this situation `dim_0` would often correspond to the latitude of the 2D variable but the depth of the 3D one. These dimensions generally have different sizes and so cannot be merged.  

Another issue which affected sub-setting with a time range was that a number of the calculated variable functions assumed that the first dimension of the data array was always depth. However, when a time range is selected the first dimension will be time. This prevented sub-setting of any variables derived from sound speed (critical depth, depth excess, sound channel axis, etc). 

I also noticed that the Navigator will try to return a time range subset for single time stamps if any timestamp other than recent is selected. This is because the `output_endtime` variable is set to the most recent time by default, and only updated when a time range is selected. 

Rather than simply rename the dimensions of the DataArray returned by `calculated.isel()` I decided to create a new object. This might require a little more code but ensures that's the data is the expected shape with correctly named dimensions, and we can also include the proper coordinates that go with the dimensions. This fix allows us to subset multiple variables and fixes issue #897.

To correct the second issue some of the calculated functions now swap the first two dimensions of the data arrays so that depth is in the 0th position. This is required since the subroutines used for these calculations assume that depths is 0th which is not the case when the variable spans multiple time stamps. It would of been nice to programmatically determine which dimension was depth and pass that to these functions however since this varies quite a lot by request there wasn't a dependable way to do this. 

Lastly, to ensure that users only get subsets for the timestamp requested we can just check to see if the 'Time Range' check box is selected. If not we can just set 'output_starttime' and 'output_endtime' to the same value for the query. 

I've tested this work for the datasets and variables listed in the 'Subset Testing' sheet I started a while back (found [here](https://docs.google.com/spreadsheets/d/1yNJT0nHrWS4-R7TxMf2XSgvIDqMpKQjHYvYWnB17854/edit?usp=sharingl)). This seems to clear up most of the issues we've been experiencing with the Subsetting function lately but not all. This functionality still doesn't work for some variables such as PSSC and some datasets including CIOPS but these issues are outside of the scope of this PR. 

## Checks
- [X] I ran unit tests.

### DIFF
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -180,4 +180,15 @@ class CalculatedArray():
         for d in self.dims:
             key.append(keys[d])
 
-        return self[tuple(key)]
+        selection = self[tuple(key)]
+        # check to see that dimensions are named, if not recreate DataArray w/ correct dims and coords
+        if selection.dims != tuple(keys.keys()):   
+            shape = [k.stop - k.start for k in keys.values()]
+            selection = xr.DataArray(
+                            data = np.reshape(selection.data, shape), 
+                            dims = keys.keys(),
+                            coords = {str(k) : self._parent.coords[k][v] for k,v in keys.items()},
+                            attrs = self.attrs
+                        )
+
+        return selection
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -132,7 +132,7 @@ class CalculatedArray():
         else: 
             dims = self._dims
 
-        key = [k for k in key if type(k) is slice] #remove any non-slice elements 
+        key = [filter(lambda k: type(k) is slice, key)]
         return xr.DataArray(
                     data = np.reshape(data_array.data, [k.stop - k.start for k in key]), 
                     dims = dims,
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -127,7 +127,18 @@ class CalculatedArray():
         data_array = self._parser.parse(
             self._expression, self._parent, key, self._dims)
 
-        return xr.DataArray(data_array, attrs=self.attrs)
+        if hasattr(data_array, 'dims'): 
+            dims = data_array.dims
+        else: 
+            dims = self._dims
+
+        key = [k for k in key if type(k) is slice] #remove any non-slice elements 
+        return xr.DataArray(
+                    data = np.reshape(data_array.data, [k.stop - k.start for k in key]), 
+                    dims = dims,
+                    coords = {str(d) : self._parent.coords[d][k] for d,k in zip(dims,key)},
+                    attrs = self.attrs
+                )
 
     def __calculate_var_shape(self) -> tuple:
         # Determine shape of calculated variable based on its
@@ -180,15 +191,4 @@ class CalculatedArray():
         for d in self.dims:
             key.append(keys[d])
 
-        selection = self[tuple(key)]
-        # check to see that dimensions are named, if not recreate DataArray w/ correct dims and coords
-        if selection.dims != tuple(keys.keys()):   
-            shape = [k.stop - k.start for k in keys.values()]
-            selection = xr.DataArray(
-                            data = np.reshape(selection.data, shape), 
-                            dims = keys.keys(),
-                            coords = {str(k) : self._parent.coords[k][v] for k,v in keys.items()},
-                            attrs = self.attrs
-                        )
-
-        return selection
+        return self[tuple(key)]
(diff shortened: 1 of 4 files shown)


data/calculated_parser/functions.py
```diff
@@ -224,7 +224,9 @@ def sspeed(depth: Union[np.ndarray, xr.Variable],
 
     if salinity.shape != press.shape:
         # pad array shape to match otherwise seawater freaks out
-        press = press[..., np.newaxis]
+        for ax, val in enumerate(salinity.shape):  
+            if ax > press.ndim - 1 or press.shape[ax] != val:
+                press = np.expand_dims(press, axis=ax)
 
     speed = seawater.svel(salinity, temperature, press)
```
JustinElms, line 231: Not 100% happy with how I did this but it works. The old method added an axis to the end of the array but when we're requesting data over a time range we may also need to pad the beginning so that `press` can be broadcast against `temperature` and `salinity`. For example, if sub-setting GIOPS sound speed salinity shape may be `(3, 50, 3, 12)` and press shape `(50, 3)`. Here we'd us expand dims to give press shape `(1, 50, 3, 1)` which is broadcastable.

nsoontie:
@JustinElms thanks for documenting these changes and potential improvements. 

It would be nice if there was a better way of checking if the arrays have a time dimension...

@htmlboss Have you had a chance to review? Any comments?

data/calculated.py
```diff
@@ -180,4 +180,15 @@ def isel(self, **kwargs):
...
+        selection = self[tuple(key)]
+        # check to see that dimensions are named, if not recreate DataArray w/ correct dims and coords
+        if selection.dims != tuple(keys.keys()):   
+            shape = [k.stop - k.start for k in keys.values()]
+            selection = xr.DataArray(
+                            data = np.reshape(selection.data, shape), 
+                            dims = keys.keys(),
+                            coords = {str(k) : self._parent.coords[k][v] for k,v in keys.items()},
+                            attrs = self.attrs
+                        )
```
htmlboss, line 192: Since we're already calling `__getitem__` via `self[tuple(key)]`, is it possible to stick all of this logic into that function to avoid potentially creating and then throwing away a `DataArray` object?

data/calculated_parser/functions.py
htmlboss: yeah I see what you're doing here and I can't think of a cleaner way of doing it. The benefit of your method is the manipulation of empty axes which is very lean on memory (as opposed to filling the empty space with 0's).

Frankly I'd put your giops example scenario into the code comment. It instantly enabled me to understand what you're solving for here. Not everyone's first instinct is to pull up the git blame lol

JustinElms: Fair! haha. Just added a comment to explain this a bit.

data/calculated.py
JustinElms: That makes way more sense than what I did here. I was able to move it into `__getitem__` but I needed to add some logic to check how the calculated array was being indexed but I think this is a better approach overall.

JustinElms:
@htmlboss @dwayne-hart Are there any other changes that need to be made here? If not, can we merge this PR so it can be tested on staging?

data/calculated.py
```diff
@@ -127,7 +127,18 @@ def __getitem__(self, key: str) -> xr.DataArray:
         data_array = self._parser.parse(
             self._expression, self._parent, key, self._dims)
 
-        return xr.DataArray(data_array, attrs=self.attrs)
+        if hasattr(data_array, 'dims'): 
+            dims = data_array.dims
+        else: 
+            dims = self._dims
+
+        key = [k for k in key if type(k) is slice] #remove any non-slice elements
```
htmlboss, line 135: String comprehensions can get a bit hairy for understanding so just use `filter` so it's self-documenting what the general operation is doing.
```suggestion
        key = [filter(lambda k: type(k) is slice, key)]
```

htmlboss (dismissed):
One suggestion in terms of code style, but :ship:

PR #925, opened 2021-12-04
Fix reshaping of calculated variables 

JustinElms:
## Background
In PR #924 we modified the `__getitem__` function in `calculated.py` so that it will return xarray DataArray object for any calculated variable. In creating this object the `data_array` is reshaped based on the `key` parameter so that it's shape matches the requested dimensions and they can be named accordingly. This parameter is generally a tuple of slice objects except on some occasions where a 2D selection is requested such as during tile generation. In this scenario the tuple contains an int for the depth level which needs to be removed hence the use of `filter`. The filter approach works however it was originally implemented as ` key = [filter(lambda k: type(k) is slice, key)]` which returns an array of filter objects which cannot be used in the `reshape` call. Removing the brackets in this line and creating a `arr_shape` variable with the needed dimensions restores the expected functionality. 

The square brackets were left over from a previous approach that built a list from the slices instead of using a filter object. This was a last minute change that as missed during testing. Removing the brackets fixed the issue.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/bin/launch-web-service.sh
+++ b/bin/launch-web-service.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+CODE_VER=$(git status | head -1 | awk '{print $NF}')
+$(sed -i "s/CODE_VERSION/$CODE_VER/g" oceannavigator/frontend/index.html)
+
 HOST_IP=$(hostname -I | awk '{print $1}')
 
 PORT=5000
--- a/bin/runserver.sh
+++ b/bin/runserver.sh
@@ -7,22 +7,4 @@ WORKER_THREADS=1
 
 [[ "$1" == "" ]] && PORT=5000 || PORT=$1
 
-if [ ! -e /usr/bin/nproc ] ; then
-
-    if [ ! -e /usr/bin/bc ] ; then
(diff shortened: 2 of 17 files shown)


JustinElms:
This fix works for all the tested datasets except for CIOPS - we now get a value error when subsetting a single variable:
`ValueError: dimensions ['time_instant'] do not exist`
and a merge error for when subsetting multiple variables: 
`xarray.core.merge.MergeError: conflicting values for variable 'nav_lat' on objects to be combined`

The first error occurs because CIOPS datasets appear to have 'time_instant' and 'time_coordinate' coordinates. 'time_coordinate' is the specified dimension in the datasetconfig.json temperature equation but the line `time_var = find_variable("time", list(self.dataset.variables.keys()))` in netcdf_data.py grabs 'time_instant' causing the value error further down the stream. We need a way to ensure that time_var returns the right value. 

The second error is due to the staggered grid used by CIOPS models. This means that the lat and lon coordinates are variable dependent which causes the merge error, similar to the issue we encountered when trying to import the PSJ model data.

Both of these issues need to be addressed but since they're separate, dataset specific problems I think they should be addressed in their own PR's.

PR #934, opened 2021-12-24
Remove erroneous call to `addControl` in Map component

htmlboss:
## Background

Artifact from a messed up rebase that causes a white screen

The scale viewer is still being manually added on line 679 which is the intended behaviour.

Restores previous functionality. Thanks Justin for pointing this out!

Only way.

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map.jsx
@@ -231,7 +231,7 @@ export default class Map extends React.PureComponent {
         });
       }
     };
-    this.map.addControl(this.scaleViewer);
+    
     this.obsDrawSource = new olsource.Vector({
       features: [],
     });


oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -231,7 +231,7 @@ export default class Map extends React.PureComponent {
         });
       }
     };
-    this.map.addControl(this.scaleViewer);
```
htmlboss, line 234: fails because map is undefined because we haven't created it yet lol. Line ~679 is the correct placement https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/9c582c0d90746ed28755b944ec59b9d061fd2c4c/oceannavigator/frontend/src/components/Map.jsx#L679

JustinElms (approved):

PR #937, opened 2022-01-04
Implement Go button

htmlboss:
## Background

At long last...the go button

* The secret is to only update the parent components when the go button is clicked..so we just need to move that code around a bit.

* The work in #926 made this change super easy to make

## Screenshot(s)

image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


dwayne-hart (approved):

htmlboss:
Tested on https://go.oceansdata.net/ and nothing appears broken

Issue #950, opened 2022-01-21
`DATASET_DEFAULTS` time values break dataset comparison


JustinElms:
**Describe the bug**
This issue is related to the defaults.js file added in #926. This file includes a `DATASET_DEFAULTS` (in defaults.js) parameter which specifies the default options to use when the Navigator is started. Among these options we specify default values time = -1 and starttime = -2 but one the DatasetSelector retrieve the timestamps available for the dataset these values are replaced with actual timestamps. The dataset used for comparisons, `dataset_1`, also uses the defaults listed in this file, however, the time and starttime values do not get updated. So when 'Show Comparison' is selected, the query is sent with -1 instead of a timestamp which prevents any data from being queried. 

**To Reproduce**
Click 'Show Comparison' at any point when using the Navigator. The API query will fail in any scenario since the dataset time doesn't get updated.  This prevents plotting of comparison data, and tile generation in the right map window when visible. 

**Expected behavior**
The user should see the comparison data requested. 

**Screenshots**

**Additional context**
I also noticed a bunch of queries with timestamp = -1 getting sent when I start the Navigator. My guess is that the Navigator is making these requests before the timestamp gets updated to a real value.

htmlboss:
>Additional context
I also noticed a bunch of queries with timestamp = -1 getting sent when I start the Navigator. My guess is that the Navigator is making these requests before the timestamp gets updated to a real value.

You're correct with that. To solve that my thinking is to simply add an optional `latest` argument to the `/timestamps` endpoint so we can query that when the Navigator first loads. Also, it's a nice helper method without having to load all the timestamps. I don't think that's required to resolve this issue, but if I remember I can add that too.

PR #952, opened 2022-01-23
Fix `id` of right map dataset selector

htmlboss:
Related to #950 

This may resolve the issue of the compare mode not updating, given in `OceanNavigator.jsx` we expect the id to be either `dataset_0` or `dataset_1`.

Can't test this locally since my container is way out of date with the package updates.

## Checks
- [ ] I ran unit tests.
- [ ] I've tested the relevant changes from a user POV.

_Hint_ To run all python unit tests run the following command from the root repository directory:
```sh
bash run_python_tests.sh
```

PR #954, opened 2022-01-25
prefetch times for DATASET_DEFAULTS

JustinElms:
## Background
Continuing work from PR #952 to fix issue #950.

When the OceanNavigator component is created we initially set the time and startttime values of dataset_0 and dataset _1 to -1 and -2 as provided by DATASET_DEFAULTS. As described in 950, these values are not updated for dataset_1, preventing users from using the compare datasets function.  This change requests these values before the component window renders in the componentWillMount method. 

I initially added an optional 'latest' argument to the `/timestamps` so we could request just the most recent timestamp but the compare function for the Hovmoller plots still didn't work so I switched to this approach. This way we get both the start and end times before users can interact with the Navigator. There might be a better way but it works though there are still some requests to tiles with the `-1` timestamp.

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -162,7 +162,7 @@ class OceanNavigator extends React.Component {
   }
 
   componentWillMount() {
-    GetTimestampsPromise(DATASET_DEFAULTS.dataset, DATASET_DEFAULTS.variable, true).then(timeResult => {
+    GetTimestampsPromise(DATASET_DEFAULTS.dataset, DATASET_DEFAULTS.variable).then(timeResult => {
       const defaults = {...DATASET_DEFAULTS};
       defaults.starttime = timeResult.data[0].id;
       defaults.time = timeResult.data[timeResult.data.length - 1].id;
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -164,7 +164,8 @@ class OceanNavigator extends React.Component {
   componentWillMount() {
     GetTimestampsPromise(DATASET_DEFAULTS.dataset, DATASET_DEFAULTS.variable, true).then(timeResult => {
       var defaults = {...DATASET_DEFAULTS};
-      defaults.time = timeResult.data[0].id;
+      defaults.starttime = timeResult.data[0].id;
+      defaults.time = timeResult.data[timeResult.data.length - 1].id;
       this.setState({ ...defaults,
                       dataset_1 : {...defaults}
                     });
--- a/oceannavigator/frontend/src/components/OceanNavigator.jsx
+++ b/oceannavigator/frontend/src/components/OceanNavigator.jsx
@@ -18,6 +18,8 @@ import { DATASET_DEFAULTS, DEFAULT_OPTIONS } from "./Defaults.js";
 
 import { withTranslation } from "react-i18next";
 
+import { GetTimestampsPromise } from "../remote/OceanNavigator.js";
+
 const stringify = require("fast-stable-stringify");
 
 function formatLatLon(latitude, longitude) {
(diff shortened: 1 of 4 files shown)


oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -159,6 +161,17 @@ class OceanNavigator extends React.Component {
     this.setState({options});
   }
 
+  componentWillMount() {
+    GetTimestampsPromise(DATASET_DEFAULTS.dataset, DATASET_DEFAULTS.variable, true).then(timeResult => {
+      var defaults = {...DATASET_DEFAULTS};
```
htmlboss, line 166:
```suggestion
      const defaults = {...DATASET_DEFAULTS};
```

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -167,7 +180,7 @@ class OceanNavigator extends React.Component {
     }
 
     if (key === "dataset_1") {
-      this.setState(value);
+      this.setState({dataset_1: value});
```
htmlboss, line 183: Great catch 🤦

htmlboss:
>I initially added an optional 'latest' argument to the /timestamps so we could request just the most recent timestamp but the compare function for the Hovmoller plots still didn't work so I switched to this approach. 

Another idea would be to add a `earliest` argument in addition to `latest` 🤷 .

We're caching these requests in the browser now anyway so multiple calls will be cheap so I think what you've got now is fine!

oceannavigator/frontend/src/components/OceanNavigator.jsx
JustinElms: All good, took an embarrassingly long time to realize why the right map options were updating the left map haha

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -159,6 +161,17 @@ class OceanNavigator extends React.Component {
     this.setState({options});
   }
 
+  componentWillMount() {
+    GetTimestampsPromise(DATASET_DEFAULTS.dataset, DATASET_DEFAULTS.variable, true).then(timeResult => {
```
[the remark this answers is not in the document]
JustinElms: Yup, totally missed that!

JustinElms:
> > I initially added an optional 'latest' argument to the /timestamps so we could request just the most recent timestamp but the compare function for the Hovmoller plots still didn't work so I switched to this approach.
> 
> Another idea would be to add a `earliest` argument in addition to `latest` 🤷 .
> 
> We're caching these requests in the browser now anyway so multiple calls will be cheap so I think what you've got now is fine!

Thanks! I agree, this is fine for now but I'm sure there's a nicer way to achieve the same result. An 'earliest' argument could definitely work here too. If that's something you think would be helpful I can revive the work I initially started here add these options to the timestamps endpoint.

oceannavigator/frontend/src/components/OceanNavigator.jsx
htmlboss: Don't sweat it lol, I guarantee it would have taken me just as long to see that.

htmlboss:
> > > I initially added an optional 'latest' argument to the /timestamps so we could request just the most recent timestamp but the compare function for the Hovmoller plots still didn't work so I switched to this approach.
> > 
> > 
> > Another idea would be to add a `earliest` argument in addition to `latest` 🤷 .
> > We're caching these requests in the browser now anyway so multiple calls will be cheap so I think what you've got now is fine!
> 
> Thanks! I agree, this is fine for now but I'm sure there's a nicer way to achieve the same result. An 'earliest' argument could definitely work here too. If that's something you think would be helpful I can revive the work I initially started here add these options to the timestamps endpoint.

I mean it'd be more semantically correct, but wouldn't give us any real performance benefit due to the fast caching being way quicker than yet another api call. 🤷

htmlboss (approved):
Great use of `componentWillMount`!

PR #961, opened 2022-02-02
API: Enable /data to also return 2D variables

htmlboss:
## Background

Fixes a bug where a 2D dataset variable is incorrectly sliced...it has no depth

Seems like the most logical approach.

* This will allow us to expand our offerings of velocity arrows to more datasets. PRs for those will be put up after this is merged.

## Screenshot(s)

Tested on 2D CCG RIOPS which was previously crashing
image

3D still works:
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/routes/api_v1_0.py
+++ b/routes/api_v1_0.py
@@ -358,18 +358,23 @@ def get_data_v1_0():
         lat_slice = slice(0, lat_var.size, 4)
         lon_slice = slice(0, lon_var.size, 4)
 
-        time_index = ds.nc_data.timestamp_to_time_index(result['time'])
+        time_index = ds.nc_data.timestamp_to_time_index(result['time'])     
         
-        data = ds \
-                .nc_data \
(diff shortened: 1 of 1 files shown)


routes/api_v1_0.py
```diff
@@ -358,18 +358,23 @@ def get_data_v1_0():
         lat_slice = slice(0, lat_var.size, 4)
         lon_slice = slice(0, lon_var.size, 4)
 
-        time_index = ds.nc_data.timestamp_to_time_index(result['time'])
-        
-        data = ds \
-                .nc_data \
-                .get_dataset_variable(result['variable'])[time_index, result['depth'], lat_slice, lon_slice]
```
htmlboss, line 365: Before, we were slicing all variables the same way...even if it was a 2D variable, we'd assume it has a depth dimension which is totally false lol.

prodyut1978: Thank you Nabil

dwayne-hart (approved):

PR #966, opened 2022-02-08
Issue 923

### LINKED ISSUE #923: missing depth drop down
Several 3D datasets do not have a 'depth' drop down (several CMEMS datasets including dataset 19)

image

Possibly associated with this issue is another:
Dataset 15.CMEMS... has a depth variable but only has '0' or 'Bottom' in the dropdown.

prodyut1978:
## Background

Resolves #923 

3D datasets do not have a 'depth' drop down when changing the dataset from 2D variables to 3D variables
In DatasetSelector.js file, depthSelector component modified with an additional filter :{ !this.state.two_dimensional} which only accept depth for 3D dataset.  

Thoroughly checked all the option including point plot and  Line plot. 

## Checks
- [ ] I ran unit tests.
- [yes] I've tested the relevant changes from a user POV.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/AreaWindow.jsx
```diff
@@ -597,7 +597,7 @@ class AreaWindow extends React.Component {
         plot_query.neighbours = this.props.options.interpNeighbours;
         plot_query.plotTitle = this.state.plotTitle;
         if (this.state.dataset_compare) {
-          plot_query.compare_to = this.props.dataset_1;
+          plot_query.compare_to = this.state.dataset_1;
```
JustinElms, line 600:
```suggestion
          plot_query.compare_to = this.props.dataset_1;
```

I think this was left over from the hack we used to get around the compare datasets issue we were having.

oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -244,6 +246,21 @@ class DatasetSelector extends React.Component {
...
+        });
+        GetDepthsPromise(this.state.dataset, value).then(depthResult => {
+          this.setState({
+            datasetDepths: depthResult.data,
+            depth: 0,
+            loading: false,
+            loadingPercent: 0
+          })
+        });
+      }
```
htmlboss, line 263: I'd suggest moving this into the `changeVariable` function, since this is a side effect of changing a variable.

oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -183,6 +200,7 @@ class DatasetSelector extends React.Component {
       newState = {
         variable: newVariable,
         variable_scale: variable.scale,
+        two_dimensional : this.state.datasetVariables.find(v => v.id === newVariable).two_dimensional,
```
htmlboss, line 203: We've already found the object corresponding to the `newVariable` on line 198. No need to perform the search twice.

```suggestion
        two_dimensional: variable.two_dimensional,
```

oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -373,7 +391,7 @@ class DatasetSelector extends React.Component {
 
     let depthSelector = null;
     // eslint-disable-next-line max-len
-    if (this.props.showDepthSelector && this.state.datasetDepths && this.state.datasetDepths.length > 0 && !this.state.loading) { 
+    if (this.props.showDepthSelector && this.state.datasetDepths && this.state.datasetDepths.length > 0 && !this.state.loading && !this.state.two_dimensional) {
```
htmlboss, line 394: Could we make this new state variable a little more descriptive? Like `variable_two_dimensional` for example.

PR #969, opened 2022-02-12
WIP Dump kidsysco time picker, jquery ui, and bower

htmlboss:
## Background

* Dumps `bower`, `jquery-ui`, and `jquery-ui-month-picker` as dependencies in favour of the already-installed `react-datepicker`.
* Updates `node-sass` to latest to remove Python 2 dependency.
* Deletes `ContinuousTimePicker` since it's now redundant.

* We already have been using the `react-datepicker` package for a couple of years. It appears to offer identical functionality to the kidsysco one so I opted to just update it to latest and use it.

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -100,6 +102,12 @@ class DatasetSelector extends React.Component {
         this.setState({ loadingPercent: 75 });
 
         const timeData = timeResult.data;
+
+        // Map timestamps to datetimes (2275776000 => 2022-02-12T00:00:00+00:00)
+        const ncTimestampDateTimeMap = new Map(data.map(i => [i.id, i.value]));
+        
+        // Map datetimes to timestamps (2022-02-12T00:00:00+00:00 => 2275776000)
+        const dateTimeNCTimestampMap = new Map(data.map(i => [i.value, i.id]));
```
htmlboss, line 110: image

JustinElms:
New timepicker added and JS packages updated in v9.0.

PR #970, opened 2022-02-21
Jjyy igoss

prodyut1978:
## Background
Implementation of a new feature of file export, named  IGOSS-JJYY file format from plot data

To Implement IGOSS-JJYY file export format data from the plot data for single point selection and multiple points import (CSV import)

Need to extend this feature for area selection 

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.
- [yes ] I've tested the relevant changes from a user POV.

```sh
bash run_python_tests.sh
```


htmlboss:
could you provide an explanation of this file format please? (expected values, headers, etc.) I'll help with reviewing the code.

prodyut1978:
> could you provide an explanation of this file format please? (expected values, headers, etc.) It'll help with reviewing the code.
Hi Nabil, I update the PR with a detail background. Please let me know if you need any other details

plotting/plotter.py
```diff
@@ -321,6 +323,77 @@ def odv_ascii(self, cruise="", variables=[], variable_units=[],
...
+            latitude = p[0]
+            longitude = p[1]
+            if latitude > 0. and longitude > 0. :
+                quad_id = 1
+            elif latitude < 0. and longitude > 0. :
+                quad_id = 3
+            elif latitude < 0. and longitude < 0. :
+                quad_id = 5
+            elif latitude > 0. and longitude < 0. :
+                quad_id = 7
```
htmlboss, line 340: What do these `quad_id` mean, and what does the 1, 3, 5, 7 represent?

htmlboss, line 340: What are these quad_id's, and what does the 1, 3, 5, 7 represent?

plotting/plotter.py
```diff
@@ -321,6 +323,77 @@ def odv_ascii(self, cruise="", variables=[], variable_units=[],
...
+        """
+        igoss_data   = self.format_igoss_data(data, depths)
+        igoss_header = self.format_igoss_header(points, time)
+        igoss_data = np.append(igoss_header, igoss_data, axis=2)
+        igoss_data = np.reshape(igoss_data, (int(igoss_data.size/7),7))
+        igoss_data = igoss_data[~np.all(igoss_data == '', axis=1)]
+        with contextlib.closing(StringIO()) as buf:
+            for r in igoss_data:
+                buf.write(" ".join(r) + '\n')
+            return (buf.getvalue(), self.mime, self.filename)
```
htmlboss, line 396: From the example igoss file you included in the pr summary, it's basically a fancy CSV file with spaces as delimiters instead of commas.

I think much of this array manipulation could be simplified by directly using the python `csv` module.

Example article on using the `csv` module: https://www.programiz.com/python-programming/writing-csv-files
docs: https://docs.python.org/3/library/csv.html

plotting/plotter.py
JustinElms: This format doesn't allow us to use negative coordinates or cardinal directions for the lat and lon values. Instead, it divides the globe into four quadrants:
 1 North East
 3 South East
 5 South West
 7 North West
The operator and software ingesting this data can use this number to determine where the coordinates are.

htmlboss: image

JustinElms: Yup haha.

plotting/plotter.py
JustinElms: It looks like this bit of code is based on the proof of concept I wrote for this feature last summer. Looking back, I think I was going through the numpy docs and using every function could haha. The reason I took this approach is because DND needs to be able to get jjyy data for multiple points simultaneously in one output file. The vectorized approach is horrible to read but it seems to scale well when multiple points were requested since it formats all the data at once. 

If we were to incorporate some of the csv module's functionality here what approach would you suggest?

JustinElms:
@htmlboss Any other thoughts or suggestions on this before we merge it?

htmlboss (approved):
Disgusting file format

dwayne-hart:
@prodyut1978 and @JustinElms this work needs to have conflicts in both plotting/plotter.py and plotting/utils.py resolved. We should then be able to merge it into the mainline for the next release.

JustinElms:
Last I heard the software that depended on this feature has become unavailable so this isn't needed yet. Not sure that this format has any use outside of that so closing for now.

PR #972, opened 2022-02-22
Fix calculated variable unit tests

JustinElms:
## Background

The changes made to the `CalculatedArray` class in PRs #924 and #925 last fall broke many of the unit tests in  _test_calculated_data.py_. The main cause of these broken tests was that the code used to assign coordinates to the `CalculatedArray` objects was not intended to handle scalar indices. Scalar indices cause issues when grabbing the coordinates for the `CalculatedArray object because slicing coordinates with them returns an xarray coordinate object with zero length which cannot be assigned to a non-empty DataArray. To resolve this I added the `_format_key()` helper function which ensures that all of the indices passed to `__getitem__()` are iterable. This allows us to assign coordinates to these dimensions without issue. 

Other tests failed because DataArray wasn't returning a numpy array in some cases. To make sure that the parser always returns a numpy array we now initially set `self.result = np.array(np.nan)` then check that the parsed function result is an array before reassigning. 

I initially started by modifying the tests to work with the new code but once I realized that most of these issues were related to the scalar indices I thought it would make more sense to make the few changes needed to make the `CalculatedArray.__getitem__()` more robust and fix the unit test issues.

## Checks
- [X] I ran unit tests.

### DIFF
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -112,6 +112,7 @@ class CalculatedArray():
         self._attrs: dict = attrs
         self._db_url: str = db_url
         self._shape: tuple = self.__calculate_var_shape()
+        self._coords: dict = {} 
 
         # This is a bit odd, but necessary. We run the expression through the
         # lexer so that the lexer variables get populated. This way we know the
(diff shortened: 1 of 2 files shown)


data/calculated.py
```diff
@@ -127,23 +127,37 @@ def __getitem__(self, key: str) -> xr.DataArray:
...
+
+    def _format_key(self, key):
+        # ensure key and its elements are iterable. This ensures 
+        # coords have length > 0
+        try:
+            key = [[k] if isinstance(k, int) else k for k in key]
+        except TypeError:
+            key = [key]
+        return key
+
```
htmlboss, line 160: Could you add some type hints to the return type and function params?

JustinElms: I definitely should have included return typing but I was hesitant to specify a type for the key argument. The hint on 124 suggests that key should be a string but it's pass as a list/int/slice/tuple throughout the code and unit tests. Any suggestion on how to handle that without using a bunch of nested unions?

htmlboss: Well it is possible but that hint declaration would be super gross. Forget it then 😂

htmlboss (approved):

Issue #39, opened 2017-07-27
Add help button to menu.


htmlboss:
* This would redirect to the Github pages.

htmlboss:
After some discussion, it was decided to have a pop-up modal that has an iframe linked to the Github Pages site (to prevent more redirections).

PR #976, opened 2022-02-24
Add Class4 OLA files to the Navigator

JustinElms:
## Background
A new type of class4 dataset will be added to the Navigator. While similar to the class4 dataset currently available, this new product will compare observed data with RIOPS model forecasts. The currently class4 dataset will be renamed as OceanPredict, and the new data as RIOPS Assimilated Observations. The most notable difference between the two is that the RIOPS Assimilated Observations data does not contain additional models or climatology.

Following suggestions from ECCC, I added two buttons to the Class4 dropdown on the map toolbar: OceanPredict and RIOPS Assimilated Observations. Clicking on either button will display a calendar indicating which days data is available as before. Clicking on any of the highlighted dates will populate the global map with points where data is available from the selected dataset. 

Since the general structure of both datasets' files is the same the logic remains fairly close to the original implementation. `CLASS4_OLA_FNAME_PATTERN` and `CLASS4_OLA_PATH` environment variables were added to oceannavigator.cfg to provide access to the new datasets. If statements and switches are used in a number of places to ensure the nc files are pulled from correct class4 directory. Some of the front end components were duplicated to accommodate having two similar dataset but I'm working to reduce this where possible. 

Example of environment variables in oceannavigator.cfg:
```
CLASS4_FNAME_PATTERN = "/%s/%s/%s.nc"
CLASS4_OP_PATH = "/data/class4/"
CLASS4_RAO_PATH = "/data/class4-ola/"
```
The file name patterns have been changed since the previous pattern wasn't compatible with our data stores directory structure and symlinks were needed to help the Navigator find the NC files. The two Class4 types will have the same directory structure so only one `CLASS4_FNAME_PATTERN` variable is needed.

## Screenshot(s)
class4_dropdown
class4_calendar
class4_op

## Checks
- [x] I ran unit tests.
- [x] I've formatted my Python code using `black`.


oceannavigator/frontend/src/components/MapToolbar.jsx
```diff
@@ -74,36 +76,71 @@ class MapToolbar extends React.Component {
...
+        beforeShowDay: this.beforeShowDay_op.bind(this),
         regional: this.props.i18n.language,
         onSelect: function(text, picker) {
-          this.props.action("show", "class4", this.state.class4Files[text]);
-          this.class4Picker.hide();
+          this.props.action("show", "class4", this.state.class4Files_op[text], "class4_op");
+          this.class4OpPicker.hide();
         }.bind(this),
         defaultDate: this.state.class4Current,
       });
```
JustinElms, line 96: Trying to find a way to simplify this section so that we only need one buttonhandler method. I'd like to do something like this but I'm having trouble finding away to get references to `this.class4OpDiv` and `this.class4RoaDiv` working here. Any suggestions @htmlboss @prodyut1978 ?
```suggestion
class4ButtonHandler(class4type) {
    const button = null;
    const div = null;
    const files = null;
    switch (class4type) {
      case "class4_op":
        button = $(ReactDOM.findDOMNode(this.class4OpButton));
        div = this.class4OpDiv;  // doesn't work below
        files = this.state.class4Files_op[text];
        break;
      case "class4_rao":
        button = $(ReactDOM.findDOMNode(this.class4RoaButton));
        div = this.class4RaoDiv;
        files = this.state.class4Files_rao[text];
        break;
    }
    if (this.class4Picker && this.class4Picker.is(":visible")) {
      this.class4Picker.hide();
    } else if (!this.class4Picker) {
      if (this.class4Picker !== null) {
        this.class4Picker = null;
      }
      this.class4Picker = $(div).datepicker({
        dateFormat: "yy-mm-dd",
        beforeShowDay: this.beforeShowDay_op.bind(this),
        regional: this.props.i18n.language,
        onSelect: function(text, picker) {
          this.props.action("show", "class4", files, class4type);
          this.class4Picker.hide();
        }.bind(this),
        defaultDate: this.state.class4Current,
      });

      $(div).css("left", button.offset().left + "px");
    } else {
      this.class4Picker.show();
    }
    this.forceUpdate();
  }
```
```

**Comment by JustinElms** [comment]
@htmlboss It looks like we might be waiting a little bit for the new Class4 OLA files to become available so this PR is going to be on hold for now. Feel free to continue with PR #969 if you'd like and I can update this accordingly.

File: oceannavigator/frontend/src/components/Class4Window.jsx [hunk 2]
```diff
@@ -42,7 +42,7 @@ class Class4Window extends React.Component {
   }
 
   onLocalUpdate(key, value) {
-    if (this._mounted) {
+    if (this._mounted && value) {
```
JustinElms, line 45: Need to check if `value` is undefined here because the /class4/models/ endpoint returns an empty list for the OLA files which JS interprets as undefined. This causes issues down the line when querying the Class4 OLA plots.

htmlboss: JS == :facepalm:

htmlboss:
Looks fine to me. Just need to fix that failing test ` KeyError: 'CLASS4_OP_PATH'`...I guess we need an update to the configs repo?

JustinElms:
> Looks fine to me. Just need to fix that failing test ` KeyError: 'CLASS4_OP_PATH'`...I guess we need an update to the configs repo?

I think so. I just opened a PR on the configs repo ([#39](https://github.com/DFO-Ocean-Navigator/configurations/pull/39)) so well see if the tests pass once that's updated.

JustinElms:
@htmlboss Looks like everything's good to go now that the configs have been updated.

htmlboss (approved):

Issue #977, opened 2022-02-25
Easier selection method for predefined areas


JustinElms:
**Priority (low, medium, high)**
Priority: low/medium

**Is your feature request related to a problem? Please describe.**
Users have reported that it can be difficult to produce an area plot when using the predefined areas available in the Navigator. Once an area is selected users need to click on the border or text of an area to plot it which isn't intuitive. Once users understand the functionality these items are still generally small and can be difficult to click on, especially when areas share a border or when there are many small areas in a region.

**Describe the solution you'd like**
Ideally, we could treat these areas like individual buttons so that clicking anywhere within the area would produce a plot. If possible, we could also highlight the area when the mouse is hovering above it like so:
![area_button](https://user-images.githubusercontent.com/79917349/155771348-dbefae72-9733-4e2e-aed6-dac667bc96b7.png)

PR #978, opened 2022-03-03
Improve UX for predefined areas

JustinElms:
## Background
From issue #977:
Users have reported that it can be difficult to produce an area plot when using the predefined areas available in the Navigator. Once an area is selected users need to click on the border or text of an area to plot it which isn't intuitive. Once users understand the functionality these items are still generally small and can be difficult to click on, especially when areas share a border or when there are many small areas in a region.

To make these areas easier to use I added a white outline to the borders and text of the areas to improve visibility and added highlighting to individual areas when the mouse hovers over them. I also made the text font larger. The user can now open area plots by clicking anywhere within the area instead of having to click on the text or borders. 

Just some small tweaks to improve visibility and functionality. A fill is needed so that the user can click anywhere within the area to open an area plot. This could be totally transparent but I though some highlighting on the selected area would make for some useful feedback. 

## Screenshot(s)
Updated UI on the left. Bottom row shows behavior for mouse hover. This image is scaled down and compressed so it's not very convincing...
areas

## Checks
- [ ] I ran unit tests.


oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -734,6 +753,44 @@ export default class Map extends React.PureComponent {
...
+                  fill: new olstyle.Fill({
+                    color: "#000000",
+                  }),
+                  stroke: new olstyle.Stroke({
+                    color: "#FFFFFF", 
+                    width: 2,
+                  }),
+                }),
+              }),  
+            ]);
```
JustinElms, line 790: It seems like overkill to have to completely redefine the style when I really just want to update the fill for the object. I couldn't find a way to do that in openlayer's documentation. Any suggestions here?

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -412,6 +412,15 @@ export default class Map extends React.PureComponent {
...
                 return [
+                  new olstyle.Style({
+                    stroke: new olstyle.Stroke({
+                      color: "#FFFFFF",
+                      width: 2,
+                    }),
+                    fill: new olstyle.Fill({
+                      color: "#FFFFFF00"
+                    })
+                  }),
```
JustinElms, line 423: It doesn't appear that openlayer's provides a method to outline polygons so as I workaround I layered two Styles with different colours and sizes to get the same effect.

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -814,25 +871,7 @@ export default class Map extends React.PureComponent {
...
-              geometry: new olgeom.Point(olproj.transform(feat.get("centroid"), "EPSG:4326", this.props.state.projection)),
-              text: new olstyle.Text({
-                text: feat.get("name"),
-                fill: new olstyle.Fill({
-                  color: "#0099ff",
-                }),
-              }),
-            }),
-          ];
-        } else {
```
JustinElms, line 835: Removed since the mouse over highlighting overrides these style changes.

oceannavigator/frontend/src/components/Map.jsx
```diff
@@ -946,6 +985,9 @@ export default class Map extends React.PureComponent {
         this.infoRequest.abort();
       }
       this.infoOverlay.setPosition(undefined);
+      if (e.selected[0].get("type") == "area") {
+        this.selectedFeatures.clear();
+      }
```
JustinElms, line 990: Clear the selectedFeatures object so that users can click on the same area to reopen the plot.

oceannavigator/frontend/src/components/Map.jsx
htmlboss: No ideas from me. Openlayers seems to have a kind of imperative API where you need to tell it everything all the time.

htmlboss (approved):
Definitely an improvement. Openlayers is sometimes a pain to work with so I'm not gonna split hairs over the copy-paste.

PR #980, opened 2022-03-04
New spec and environment files.

dwayne-hart:
Bringing in flake and black in order to implement software development best practices.


htmlboss:
you _might_ need to also bring in the `frozendict` package. booting up the navigator on master with this new spec file may result in an error if it's missing.

dwayne-hart:
Thanks. I'll gladly put that in via a separate PR. However, I did not find
the package listed in any of our comms.

htmlboss:
oops yeah that's my bad. I had it in another list which I improperly copied over.

PR #981, opened 2022-03-04
Enter multiple points

prodyut1978:
## Background
The UI  Enter Coordinates option through  Point  selection menu provides the user input point location (Latitude-Longitude) in the map and also graph plot the various dataset variables (Temp, Salinity, etc.) against the different level of depth. The user can also export the graph plot data in various format (e.g. .csv, IGOSS-JJYY, ODV)

Before the Enter coordinates option, the user only able to enter one coordinate to plot that point in the map. The new added feature now it is possible to enter multiple points in the map and also plot multiple data points together in the graph. 

Some control and restrictions is added in the button when using this feature

## Screenshot(s)

Capture1
Capture2
## Checks
- [ ] I ran unit tests.
Capture2

- [y ] I've tested the relevant changes from a user POV.

```sh
bash run_python_tests.sh
```


JustinElms:
A couple comments from a user perspective:

- I'm only seeing the first coordinate at the top of the plot window when plotting multiple ones. Can you add all of the entered coordinates to the top of the `PointWindow` component similar to what we see when uploading a csv?
coords 
- Is it possible to clear the `coordInputPanel` fields in the `EnterPoint` component after you click add? That would let the user know that the coordinates were entered, otherwise users may repeatedly add the same coordinates thinking that they haven't been entered.

Also, it looks like your JJYY code was committed to this PR and causing merge conflicts here. Can you remove these changes so that the two PR's are unique?

PR #983, opened 2022-03-05
Enter multiple Coordinates

prodyut1978:
## Background
The UI Enter Coordinates option through Point selection menu provides the user input point location (Latitude-Longitude) in the map and also graph plot the various dataset variables (Temp, Salinity, etc.) against the different level of depth. The user can also export the graph plot data in various format (e.g. .csv, IGOSS-JJYY, ODV)

Before the Enter coordinates option, the user only able to enter one coordinate to plot that point in the map. The new added feature now it is possible to enter multiple points in the map and also plot multiple data points together in the graph.

Some control and restrictions is added in the button when using this feature

## Screenshot(s)
Capture1
Capture2

## Checks
- [ ] I ran unit tests.
- [ yes] I've tested the relevant changes from a user POV.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/EnterPoint.jsx
```diff
@@ -4,8 +4,40 @@ import CoordInputPanel from "./CoordInputPanel.jsx";
...
+    }
+  }
+
+  onAdd () {
+    var newState = this.state;
+    this.setState(prevState => ({
+      coordinate: [prevState.coordinate, newState]
+    }))
+    this.props.addCoordData(newState, this.props.id); // Update Added List
+    console.log(this.state)
```
JustinElms, line 39:
``` suggestion
```

JustinElms:
Feature added in v9.0.

PR #987, opened 2022-03-10
Fix broken quiver arrows again >.>

htmlboss:
## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/__init__.py
+++ b/oceannavigator/__init__.py
@@ -39,7 +39,7 @@ def config_dask(app) -> None:
     )
 
 
-def create_app(testing: bool = False):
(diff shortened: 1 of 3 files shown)


oceannavigator/__init__.py
```diff
@@ -39,7 +39,7 @@ def config_dask(app) -> None:
     )
 
 
-def create_app(testing: bool = False):
+def create_app(testing: bool = False, dataset_config_path: str = ""):
```
htmlboss, line 42: Pass the dataset config path we want to load in here so we don't have to mock this every time in the tests! Too much mocking going on atm.

PR #988, opened 2022-03-10
Fix calculated indexing

JustinElms:
## Background
I forgot to account for the : operator when I made the changes in #972 which results in an empty slice index when passed into `__getitem__`. To avoid this I'm just taking the array shape from the coordinates since we really only need to reshape the array when it's dimensions differ from it's parents dimensions/coordinates.

Makes more sense to get the final array shape from the coordinates since we're reshaping the array to match these anyway.

## Checks
- [x] I ran unit tests.

### DIFF
--- a/data/calculated.py
+++ b/data/calculated.py
@@ -132,8 +132,8 @@ class CalculatedArray:
         key = self._format_key(key)
         coords = self._calculate_coords(key)
 
-        if data_array.ndim != len(self._dims):
-            shape = [k.stop - k.start if isinstance(k, slice) else 1 for k in key]
+        if coords and data_array.ndim != len(self._dims):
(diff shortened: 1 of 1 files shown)


htmlboss:
LOL you beat me to the punch. I was writing a test in https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/pull/987/files which you can use! Just need to add a new sample netcdf file yanked from `giops_day`.

JustinElms:
> LOL you beat me to the punch. I was writing a test in https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/pull/987/files which you can use! Just need to add a new sample netcdf file yanked from `giops_day`.

That would be great, thanks!!

PR #990, opened 2022-03-11
Create action to run python tests

htmlboss:
## Background

Automatically run python tests for PRs so we can stop asking each other if we ran tests every 3 seconds.

We can only test ubuntu using the github-hosted actions. If we want other environments, we must pursue self-hosted runners https://docs.github.com/en/actions/hosting-your-own-runners

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  push:
+    branches: [ master, ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
```
htmlboss, line 12: TODO: Also run on the ubuntu version in prod and maybe other OS's too.

PR #996, opened 2022-03-12
Remove a bunch of stubbing in python tests

htmlboss:
## Background

Instead of stubbing out a chunk of the code path, just feed Flask a datasetconfig file specifically made for testing so we can execute the entire code path (i.e. integration test) for better test coverage.

* Tweaks the `create_app` function to accept an optional `dataset_config_path` which the API tests will leverage to pass in the  existing `tests/testdata/datasetconfigpatch.json`.
* Fills in some missing data in `tests/testdata/datasetconfigpatch.json` so it's actually usable.
* Removes all stubbing in the working API tests.
* Deletes some dead tests that aren't relevant anymore.

* Try to only stub the outer boundaries of an application (e.g. network access, database access, etc) when writing integration tests so we ensure maximum test coverage.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


htmlboss:
Ready to merge

dwayne-hart (approved):

Issue #999, opened 2022-03-25
Cannot generate bathymetry tiles for polar projections


JustinElms:
**Describe the bug**
The following exception is thrown when the user switches to the polar projections in the Navigator map settings:

```
Traceback (most recent call last):
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/sentry_sdk/integrations/flask.py", line 90, in sentry_patched_wsgi_app
    return SentryWsgiMiddleware(lambda *a, **kw: old_app(self, *a, **kw))(
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/sentry_sdk/integrations/wsgi.py", line 140, in __call__
    reraise(*_capture_exception(hub))
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/sentry_sdk/_compat.py", line 54, in reraise
    raise value
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/sentry_sdk/integrations/wsgi.py", line 133, in __call__
    rv = self.app(
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/sentry_sdk/integrations/flask.py", line 90, in <lambda>
    return SentryWsgiMiddleware(lambda *a, **kw: old_app(self, *a, **kw))(
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/flask/app.py", line 2091, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/flask/app.py", line 2076, in wsgi_app
    response = self.handle_exception(e)
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/ubuntu/tools/miniconda/3/amd64/envs/navigator/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
[...]

PR #1002, opened 2022-03-31
Cannot generate bathymetry tiles for polar projections #999

prodyut1978:
## Background
Fixes issue #999 

The query was using bathymetry type instead of that selected projection type. Changing `currentBathy` to `currentProj` fixing the problem.
```
updateBathymetryLayer(currentProj, currentBathy) {
    let source = null;
    switch (currentBathy) {
      case "etopo1":
      default:
        source = new olsource.XYZ({
          url: (
            `/api/v1.0/tiles/bath/${currentProj}` +
            "/{z}/{x}/{y}.png"
          ),
          projection: currentProj,
        })
        break;
    }
```

## Screenshot(s)
Capture1
Capture2
Capture3

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map.jsx
@@ -1407,7 +1407,7 @@ export default class Map extends React.PureComponent {
       default:
         source = new olsource.XYZ({
           url: (
-            `/api/v1.0/tiles/bath/${currentBathy}` +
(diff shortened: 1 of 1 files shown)


htmlboss (approved):
is `currentBathy` being used anywhere else?

prodyut1978:
>@nabil
I didnot find any link of currentBathy for other place.

Issue #945, opened 2022-01-14
Various Basemap package issues - replace package w/ Cartopy


JustinElms:
**Describe the bug**
A number of map plotting bugs in the Navigator when plotting points and lines when the 'Show Map' option is enabled, areas (e.g. Issue #832 ) and observations seem to be related to issues with the deprecated Basemap package. Per Basemap's [documentation](https://matplotlib.org/basemap/users/intro.html#cartopy-new-management-and-eol-announcement), Basemap should be replaced with Cartopy. 

**To Reproduce**
Various plotting issues encountered when plots contain Basemap objects.

Issue #832, opened 2021-04-14
WCPS-GLS Great Lakes Coupled Atmosphere-Ocean-Ice Forecast does not display data in area plot menu


JustinElms:
**Describe the bug**
Data from the WCPS-GLS Great Lakes Coupled Atmosphere-Ocean-Ice Forecast data set is not displayed in the area plot menu map. Instead, users are presented with a grey map which encompass the area drawn. The area polygon is also missing from this map. This occurs when plotting any of the variables within the data set whether the area polygon is user drawn, its coordinates are entered, or uploaded in a csv file. See screenshot below.

**To Reproduce**
Select the 08. WCPS-GLS Great Lakes Coupled Atmosphere-Ocean-Ice Forecast - LatLon data set before adding an area polygon over either of the great lakes. This can be done by drawing, entering coordinates, or uploading a csv file containing coordinates. Then click the Plot button on the toolbar to open the plot menu.

**Expected behavior**
The Plot menu map should be populated with data from the selected variable. 

**Screenshots**
![image](https://user-images.githubusercontent.com/79917349/114752457-de63b700-9d30-11eb-9154-e08fd1f0db48.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Chrome 
 - Version:  89.0.4389.114

JustinElms:
Half solution:

It looks like this issue stems from the `fillcontinents` method of Basemap used in `map.py`. So far I haven't been able to find a way to mask the lakes to prevent `fillcontinents`  from drawing over them but I was able to come with a solution that gets us close by removing the call to this method entirely and changing the background fill colour. The resulting map looks similar but gets pretty jagged along the coastline due to the resolution of the data layer. 

![gls_plot_zoom](https://user-images.githubusercontent.com/79917349/115879770-97f31400-a424-11eb-9d09-12d44fe7494b.png)
![atl_plot](https://user-images.githubusercontent.com/79917349/115879777-99bcd780-a424-11eb-92d1-f6e4fca7b8b1.png)

PR #1005, opened 2022-04-12
Replace Basemap with Cartopy

JustinElms:
## Background
The Basemap package was deprecated in favor of Cartopy for quite some time now and has been the root of some issues related to area plots and the "show location" functionality (see Issues #832 and #945). Also, Basemap is likely to cause package dependency problems in the future if not updated. 

Unlike Basemap, which provides a basemap object to handle map plotting and related functionality, Cartopy provides CRS objects that can be passed to Matplotlib axes to create a Geoaxis object. This results in a workflow that's very similar to ordinary Matplotlib plotting, users just need to pass CRS projection objects as `projection` and `transform` keyword arguments to ensure the data is plotted correctly. CRS objects also provide coordinate transformations and other utility functions. 

- Like Basemap, Cartopy draws land and lakes in separate layers which causes the problem in Issue 832. To resolve this, I created shapefiles with these geometries merged into a single layer so that Cartopy does not fill in lakes when plotting, allowing users to view the great lakes data in an area plot. I make a low, medium, and high res version of this shapefile which can be added to the datastore. 
- Cartopy does not include the Bluemarble image used in the "Show Location" plots. I've downloaded a copy which will also be added to the datastore. 
- Polar plots no longer cover the entire pole. This was pretty time consuming so I limited them to the extent of the area drawn. 

## Screenshot(s)
cartopy_1
cartopy_2

## Checks
- [ ] I ran unit tests.
- [x] I've formatted my Python code using `black `.

### DIFF
--- a/plotting/basemap.py
+++ b/plotting/basemap.py
@@ -4,10 +4,15 @@ import pickle
 import threading
 
 import cartopy.crs as ccrs
+import cartopy.feature as cfeature
 import cartopy.io.shapereader as shpreader
 import matplotlib.pyplot as plt
 from flask import current_app
 from typing import Union
+
+import pyproj
+
+import time
 
 
 def get_resolution(height: float, width: float) -> str:
@@ -25,15 +30,15 @@ def _get_shapefile(resolution: str) -> shpreader.BasicReader:
 
     if resolution == "f":
         land_shp = shpreader.BasicReader(
-            current_app.config["SHAPE_FILE_DIR"] + "ne_10m_diff.shp"
+            current_app.config["SHAPE_FILE_DIR"] + "/ne_10m_diff.shp"
         )
     elif resolution == "i":
         land_shp = shpreader.BasicReader(
-            current_app.config["SHAPE_FILE_DIR"] + "ne_50m_diff.shp"
+            current_app.config["SHAPE_FILE_DIR"] + "/ne_50m_diff.shp"
         )
     else:
         land_shp = shpreader.BasicReader(
-            current_app.config["SHAPE_FILE_DIR"] + "ne_110m_diff.shp"
+            current_app.config["SHAPE_FILE_DIR"] + "/ne_110m_diff.shp"
         )
 
     return land_shp
@@ -42,7 +47,8 @@ def _get_shapefile(resolution: str) -> shpreader.BasicReader:
 def load_map(
     plot_proj: ccrs, extent: list, figuresize: list, dpi: int, plot_resolution: str
 ) -> Union[plt.figure, plt.axes]:
-
+    
+    pyproj.set_use_global_context(active=True)
     CACHE_DIR = current_app.config["CACHE_DIR"]
     filename = _get_filename(plot_proj.proj4_params["proj"], extent)
     filename = "".join([CACHE_DIR, "/", filename])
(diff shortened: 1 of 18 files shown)


plotting/basemap.py
```diff
@@ -1,120 +1,98 @@
...
-            return "h"  # high resolution
-        elif area_km < 1000000:
-            return "i"  # intermediate resolution
-        elif area_km < 10000000:
-            return "l"  # low resolution
-        else:
-            return "c"  # crude resolution
-
-    if _maps_cache.get(filename) is None or True:
-        filename = "".join([CACHE_DIR, "/", filename])
```
JustinElms, line 43: It looks like we were changing the filename here before adding it to the _maps_cache but that prevented us from ever retrieving it. Moving the filename change to before this check allows us to grab the plot from the cache but it always returned an empty axis so I ended up scrapping it and using the pickled plot instead.

plotting/basemap.py
```diff
@@ -1,120 +1,98 @@
...
-def load_nwatlantic() -> return_type(load_map):
-    return load_map("lcc", (55, -60), 5e6, 3.5e6)
-
-
-def load_arctic() -> return_type(load_map):
-    return load_map("npstere", (65, 0), None, None)
-
-
-def load_pacific() -> return_type(load_map):
-    return load_map("lcc", (53, -137), 2e6, 2.5e6)
```
JustinElms, line 109: I couldn't find a place in the code where these functions were used so they were dropped. If they're necessary just let me know and I'll restore them.

plotting/map.py
```diff
@@ -107,37 +111,26 @@ def __load_quiver(self) -> bool:
...
-                    "npstere",
-                    self.centroid,
-                    height,
-                    width,
-                    min(self.bounds[0], self.bounds[2]),
+                self.plot_projection = ccrs.Stereographic(
+                    central_latitude=self.centroid[0],
+                    central_longitude=self.centroid[1],
                 )
+                width_scale = 1.5
```
JustinElms, line 128: This factor was added to account for projection distortion at the poles and ensure that we select a wide enough range to display the selected area.

plotting/map.py
```diff
@@ -954,6 +982,4 @@ def find_lines(values):
                 fontsize=14,
             )
 
-        fig.tight_layout(pad=3, w_pad=4)
-
         return super(MapPlotter, self).plot(fig)
```
JustinElms, line 985: This piece is incredibly slow. It's currently taking 70+ seconds to get through the `plt.savefig()` function in plotter.py. I haven't been able to figure this out yet. Any suggestions @htmlboss?

plotting/basemap.py
htmlboss: keep em gone

plotting/map.py
htmlboss: Seems to be a known issue....any chance we can use `plt.imsave` instead of `savefig`? https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imsave.html Apparently it could be faster...

In other news, matplotlib supports interactive plots in the browser via websockets if you change the plotting backend LOL

htmlboss: Setting the rendering style to fast could help too? https://matplotlib.org/stable/users/explain/performance.html#using-the-fast-style

htmlboss: For reference: https://github.com/matplotlib/matplotlib/issues/16758

JustinElms: Thanks! Looks like there's a few ways forward I can try.

JustinElms: I didn't have much luck on the matplotlib side of things. Looks like `plt.imsave` saves arrays to images which won't really work for us here and I didn't see any improvement when setting the rendering style to fast. 

I'm thinking this might be an issue with how Cartopy handles various projections. If we change the projection from LCC to Plate Carree (the default for Cartopy) I can get the savefig time down to 12 s, more or less in line with what we were getting with basemap. The plot generation will also be much quicker. The catch is that it's an equirectangular projection so there'll be some distortion in the resulting maps. 

Here's the plot from above using Plate Carree projection:
pc_plot

htmlboss:
Any chance we could cache the `crs` object between requests in Memecached or something? Some ideas in here https://github.com/SciTools/cartopy/issues/1895

plotting/basemap.py
```diff
@@ -1,120 +1,98 @@
...
-            def do_pickle(basemap: Basemap, filename: str) -> None:
-                pickle.dump(basemap, open(filename, "wb"), -1)
+            fig = plt.figure(figsize=figuresize, dpi=dpi)
+
+            map_plot = plt.axes(projection=plot_proj, facecolor="dimgrey")
+            map_plot.set_extent(extent, crs=plot_proj)
+
+            map_plot.add_geometries(
+                land_shp.geometries(),
+                crs=ccrs.PlateCarree(),
```
htmlboss, line 64: can we cache this object between requests? Apparently this guy is stupid slow to instantiate with LCC.

JustinElms: I was trying to achieve something similar by pickling the plot after setting the extent and adding the land geometries. I'm not sure if you could effectively cache the `ccrs.LambertConformal` object since you have to specify the central lat/lon when you create it, otherwise the area plotted may be skewed depending on where it is relative to the default central lat/lon (39,-96).

JustinElms:
> Any chance we could cache the `crs` object between requests in Memcached or something? Some ideas in here [SciTools/cartopy#1895](https://github.com/SciTools/cartopy/issues/1895)
> 
> Hesitant to stand up another service but I think good use of a shared Memcached instance between our processes could benefit us greatly here while cartopy gets their stuff in gear....assuming we can serialize a crs object into memcached...might need to pickle it first?

Bringing in Memcached could be helpful for a few areas of the Navigator but I'm not sure it would be useful to cache a lcc CRS object for the reasons I mentioned above. 

I saw that issue thread earlier and there's a few good ideas there. It looks like people generally set `PYPROJ_GLOBAL_CONTEXT=ON` but I saw that option isn't threadsafe so I don't think we can use it here. 
My list of things to try currently includes:
- Bump proj up to >8.1
- Check to see if I can optimize the land features by using only the geometries intersected by the map extent and/or transforming them to the projection coordinates.
- Downgrade to v19.0 to see if that makes a difference. I don't like the idea of setting a version dependency on a library as we bring it in but that might be a good temporary fix while we work on an alternative/wait to see if the issue gets resolved on their end. 
- Use crs.PlateCarree for the meantime since it doesn't seem to have these performance issues. 
- Put this on pause and stick to Basemap until something better comes along...

htmlboss:
yeah the global context is bad news for us since we run in multiple processes and potentially across multiple containers so it's dangerous but also useless.

I'd try fiddling with the package versions to start.

Besides basemap being EOL'ed, I can't remember what was forcing us to move to cartopy? It's an open source project so we could just sponsor it or we could dedicate time to get it working for our case.

JustinElms:
I encountered a few scenarios where basemap broke some of the point/line/obs plots when the 'Show Location' option was selected but beyond that not major reason to migrate to Cartopy. It just seemed like a good time to do it since we updated most of the other dependencies. 

I'll get started with the versioning and let you know if anything improves. Thanks again for the feedback!

JustinElms:
Current time to run savefig for the area above with LCC projection is ~130s. Downgrading cartopy to 0.19.0 brings that down to ~120s. Upgrading proj to 9.0 brings it ~90s. Still terrible but a fair bit better. Unfortunately I can't get a working environment with cartopy 0.19 and proj 9.0 so we cant do both. I'll look into restricting the geometries to the plot are to see if we can get any improvements there.

dwayne-hart (approved):

Issue #1003, opened 2022-04-10
Memory profiles


htmlboss:
<details>
  <summary>`/api/v1.0/tiles/gaussian/25/10/EPSG:3857/giops_day/votemper/2281521600/0/-5,30/3/3/3.png` produces the following memory profile</summary>
  
```

Filename: /home/ubuntu/onav-cloud/Ocean-Data-Map-Project/data/__init__.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     9    185.0 MiB    185.0 MiB           1   @hashable_lru
    10                                         @profile
    11                                         def open_dataset(dataset, **kwargs):
    12                                             """Open a dataset.
    13                                         
    14                                             Creates a CalculatedData (derived from NetCDFData) instance to handle dataset file
    15                                             access and calculation layer operations.
    16                                             Then, determines the type of model the dataset is from and returns the appropriate
    17                                             Model-derived instance (Nemo, Mercator, Fvcom) with the calculation layer instance
    18                                             as an attribute.
    19                                         
    20                                             Note: the returned model object will be LRU-cached internally so frequent calls
    21                                             to open the "same" dataset will have minimal overhead.
    22                                         
    23                                             Params:
    24                                                 * dataset -- Either a DatasetConfig object, or a string URL for the dataset
    25                                         
    26                                             Optional Keyword Arguments:
    27                                                 * variable {str or list} -- String or list of strings of variable keys to be loaded
    28                                                     (e.g. 'votemper' or ['votemper', 'vosaline']).
    29                                                 * timestamp {int} -- Integer value of date/time for requested data (e.g. 2128723200).
    30                                                     When loading a range of timestamps, this argument serves as the starting time.
    31                                                 * endtime {int} -- Integer value of date/time. This argument is only used when
    32                                                     loading a range of timestamps, and should hold the ending time.
    33                                                 * nearest_timestamp {bool} -- When true, open_dataset will assume the given
    34                                                     starttime (and endtime) do not exactly correspond to a timestamp integer
    35                                                     in the dataset, and will perform a binary search to find the nearest timestamp
    36                                                     that is less-than-or-equal-to the given starttime (and endtime).
    37                                             """
    38    185.0 MiB      0.0 MiB           1       MODEL_CLASSES = {
    39    185.0 MiB      0.0 MiB           1           "mercator": Mercator,
[...]

PR #1007, opened 2022-04-20
Remove `hashable_lru` decorator from `open_dataset`

htmlboss:
## Background
See #1003 for graphs

Removes one source of unbounded memory growth. This decorator isn't even being used correctly since the function arguments are being hashed based on memory location and so we never see a cache hit.

Fastest way to stem the bleeding until we have time to properly implement a memcached-backed `open_dataset` in the future.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/data/__init__.py
+++ b/data/__init__.py
@@ -2,10 +2,8 @@ from data.calculated import CalculatedData
 from data.fvcom import Fvcom
 from data.mercator import Mercator
 from data.nemo import Nemo
-from utils.decorators import hashable_lru
 
 
-@hashable_lru
 def open_dataset(dataset, **kwargs):
     """Open a dataset.
 
--- a/oceannavigator/__init__.py
+++ b/oceannavigator/__init__.py
@@ -6,6 +6,7 @@ from sys import argv
 import dask
 import sentry_sdk
 from flask import Flask, request, send_file
+from flask.logging import default_handler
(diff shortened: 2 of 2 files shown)


htmlboss:
Something's f'ed up with the Flask log formatting setup that's causing the tests to fail. I think we're doing it in the wrong spot.  https://flask.palletsprojects.com/en/2.0.x/logging/ claims we should be setting the default formatter before the `app` is created https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/5d28a350e2aed570b6e38c9484c8efce2f707beb/oceannavigator/__init__.py#L77-L82

htmlboss:
Log formatting should be fixed in the tests now. No difference in format:
image

JustinElms (approved):
Looks good! Probably worth keeping the `hashable_lru` decorator code in utils.decorators around in case we stumble upon a way to use it properly that doesn't contribute to the memory issues.

dwayne-hart (approved):

htmlboss:
> Looks good! Probably worth keeping the `hashable_lru` decorator code in utils.decorators around in case we stumble upon a way to use it properly that doesn't contribute to the memory issues.

yah exactly. The code is more-or-less okay, just needs some tweaking

PR #1009, opened 2022-04-20
Add Git Version endpoint to API and Response Header

JustinElms:
## Background
Adds an API endpoint that returns the Git version of the Navigator's code via `/api/v1.0/gitversion`. This can be useful for testing/profiling the Navigator and comparing performance of different releases. The git version is also added to the Response Headers so that we can grab it from individual API requests.

GitPython provides a more direct approach but may leak resources. I came across other approaches which account for system environment but since the Navigator is only deployed in nix systems this approach seemed most straight forward.

## Screenshots
git_version

## Checks
- [x] I ran unit tests.

### DIFF
--- a/oceannavigator/__init__.py
+++ b/oceannavigator/__init__.py
@@ -50,6 +50,25 @@ def configure_log_formatter() -> None:
     )
 
 
+def subprocess_check_output(*args) -> str:
+    try:
+        return (
+            subprocess.check_output([*args], stderr=subprocess.STDOUT)
+            .decode("ascii")
+            .strip()
+        )
+    except UnicodeDecodeError:
+        print(
+            "Unable to decode subprocess response - try updating your current branch."
+        )
+        return ""
+    except subprocess.CalledProcessError as e:
+        print(
+            f"Exception on process - args={args} rc=#{e.returncode} output=#{e.output}"
+        )
+        return ""
+
+
 def create_app(testing: bool = False, dataset_config_path: str = ""):
     configure_log_formatter()
 
@@ -64,14 +83,8 @@ def create_app(testing: bool = False, dataset_config_path: str = ""):
     app.config.from_pyfile("oceannavigator.cfg", silent=False)
     app.config.from_envvar("OCEANNAVIGATOR_SETTINGS", silent=True)
     app.testing = testing
-    app.git_hash = (
-        subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("ascii").strip()
-    )
-    app.git_tag = (
-        subprocess.check_output(["git", "describe", "--tags", "--abbrev=0"])
-        .decode("ascii")
-        .strip()
-    )
+    app.git_hash = subprocess_check_output("git", "rev-parse", "HEAD")
+    app.git_tag = subprocess_check_output("git", "describe", "--tags", "--abbrev=0")
 
     if app.config.get("WSGI_PROFILING"):
         if not os.path.isdir("./profiler_results"):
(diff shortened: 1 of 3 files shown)


JustinElms:
@htmlboss Is this kind of what you were envisioning for the response header?

routes/api_v1_0.py
```diff
@@ -1433,9 +1439,10 @@ def after_request(response):
     header = response.headers
     # Relying on iptables to keep this safe
     header["Access-Control-Allow-Origin"] = "*"
+    header["Git-Version"] = current_app.git_version
```
htmlboss, line 1442: I suggest creating a prefix for any custom headers we define. `X-Onav` basically implies a namespace that any of our custom stuff will reside.
```suggestion
    header["X-ONav-Git-Version"] = current_app.git_version
```

dwayne-hart:
Are you able to provide report the git tag being used along with the hash code?

JustinElms:
> Are you able to provide report the git tag being used along with the hash code?

I didn't think of that but probably! I'll see what I can do

JustinElms:
@dwayne-hart Added the tag to the response header and API endpoint. It's essentially just running `git describe --tags --abbrev=0` which just returns "HEAD" when working from a branch (as seen below). When I run the code in the master branch I get 'v7.2' as expected. Also, we can provide the short hash instead of long as you see below, or both if necessary.
git_version

JustinElms:
A couple of the tests are failing because I'm not specifying a path for the subprocess. Not sure if this is really an issue but any suggestions on how to fix it if so? Looks like we need to specific a path for the git executable, maybe we could add that to oceannavigator.cfg?

oceannavigator/__init__.py
```diff
@@ -63,6 +64,14 @@ def create_app(testing: bool = False, dataset_config_path: str = ""):
     app.config.from_pyfile("oceannavigator.cfg", silent=False)
     app.config.from_envvar("OCEANNAVIGATOR_SETTINGS", silent=True)
     app.testing = testing
+    app.git_hash = (
+        subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("ascii").strip()
```
htmlboss, line 68: I'd suggest extracting these two functions into `utils` maybe and catch any exceptions that may arise otherwise the app won't start.

```py

def subprocess_check_output(*args, **kwargs) -> str:
    try:
        return subprocess
            .check_output(*args, **kwargs, stderr=subprocess.STDOUT)
            .decode("ascii")
            .strip()
    except subprocess.CalledProcessError as e:
        print(f"Exception on process - args={args} rc=#{e.returncode} output=#{e.output})
        return ""
end
```

htmlboss:
I pushed a commit to your branch that will fix the subprocess error in CI via https://stackoverflow.com/a/71721059/2231969

oceannavigator/__init__.py
JustinElms: Oh good catch. I can't put that function in `utils.misc` because it imports `DatasetConfig` from `oceannavigator` which will raise an error due to the circular imports but I add it below these calls in `oceannavigator.__init__` if that sounds good to you.

htmlboss (approved):
Nice.

The CI failure is related to cartopy from the looks of it. Justin could you take a look? Might need to stub a/some tests.
`FileNotFoundError: [Errno 2] No such file or directory: '/data/misc/shapes/cartopy_resources/bluemarble.png'`

JustinElms:
> Nice.
> 
> The CI failure is related to cartopy from the looks of it. Justin could you take a look? Might need to stub a/some tests. `FileNotFoundError: [Errno 2] No such file or directory: '/data/misc/shapes/cartopy_resources/bluemarble.png'`

No problem! This has been a pain since we merged the Cartopy work. I think we just need to add some error handling for when those resources aren't available.

PR #1015, opened 2022-05-05
Add Statistics to Plots and csv Files

JustinElms:
## Background
This work adds stats data to plot images and provides users with a way to download them. Stats have been added to all plots except for T/s Diagrams, Class4/Observations, and Hovmoller plots. For area plots stats are only calculated for the selected region, otherwise they're are calculated from whatever data is displayed in the plot - for example, a multi-point plot will have max mean min data calculated from all points. 

By calculating the statistics in the plotter function we can reduce the number of data queries and easily add these statistics to csv output. This also avoids developing a new frontend component for displaying the stats. 

## Screenshot(s)

stats1
stats2

## Checks
- [X] I ran unit tests.


dwayne-hart:
Have you made an adjustment to the graph statistics informing the end user that the statistics is for the entire viewable area and not for the user defined region?

JustinElms:
Nope, but I managed to reduce the stats calculation to the drawn area so we should be good!

plotting/hovmoller.py
```diff
@@ -326,10 +335,23 @@ def _hovmoller_plot(
...
+        )
+
         plt.xlabel(gettext("Distance (km)"))
         plt.xlim([self.distance[0], self.distance[-1]])
 
         divider = make_axes_locatable(plt.gca())
         cax = divider.append_axes("right", size="5%", pad=0.05)
         bar = plt.colorbar(c, cax=cax)
-        bar.set_label("%s (%s)" % (name, utils.mathtext(unit)))
+        bar.set_label("%s (%s)" % (name, math_unit))
```
htmlboss, line 357: Can this be f-string'ed maybe?

plotting/map.py
```diff
@@ -108,6 +108,25 @@ def __load_quiver(self) -> bool:
             and self.quiver["variable"] != "none"
         )
 
+    def __apply_poly_mask(self, data) -> np.ma.MaskedArray:
```
htmlboss, line 111:
```suggestion
    def __apply_poly_mask(self, data: np.ndarray) -> np.ma.MaskedArray:
```

htmlboss: not sure if that's the right signature but you get the idea

htmlboss:
Overall looks fine! It does feel copy-pasty in some spots. Throw it into the we need to re-structure how the plotting code is written bucket.

htmlboss (approved):

PR #1018, opened 2022-05-11
Handle missing Cartopy resources

JustinElms:
## Background
Error handling is needed for deployments where the necessary Cartopy images and shapefiles may no be available, i.e. for our  automated CI testing. If these files cannot be located, the Navigator will fall back to using Cartopy's built-in resources.

If these resources are unavilable it only makes sense to use those provided by Cartopy. This should allow GitHub actions to complete CI testing.  

## Screenshot(s)
If the bluemarble image (top) is missing plot image with Cartopy's included features (bottom)
trans1
trans2

## Checks
- [X] I ran unit tests.

### DIFF
--- a/plotting/basemap.py
+++ b/plotting/basemap.py
@@ -33,6 +33,7 @@ def _get_land_geoms(resolution: str, extent: list) -> shpreader.BasicReader:
             current_app.config["SHAPE_FILE_DIR"] + shp_file
         )
     except shpreader.shapefile.ShapefileException:
+        print(f"Could not open {shp_file}, using Cartopy feature interface.")
         land_shp = cfeature.NaturalEarthFeature("physical", "land", resolution)
 
     # crop land geometries to plot extent
--- a/plotting/basemap.py
+++ b/plotting/basemap.py
@@ -4,6 +4,7 @@ import pickle
 import threading
 
 import cartopy.crs as ccrs
+import cartopy.feature as cfeature
 import cartopy.io.shapereader as shpreader
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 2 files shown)


plotting/utils.py
```diff
@@ -164,11 +163,20 @@ def _map_plot(points, grid_loc, path=True, quiver=True):
...
+        m.imshow(
+            img,
+            origin="upper",
+            extent=(-180, 180, -90, 90),
+            transform=ccrs.PlateCarree(),
+            zorder=1,
+        )
+    except FileNotFoundError:
+        m.add_feature(cfeature.LAND)
+        m.add_feature(cfeature.OCEAN)
```
htmlboss, line 179: nit: could you please add a basic `print` statement here so we can capture exception case in our rudimentary logs?

plotting/basemap.py
```diff
@@ -16,27 +17,23 @@ def get_resolution(height: float, width: float) -> str:
...
-    else:
+    shp_file = f"/cartopy_resources/ne_{resolution}_diff.shp"
+
+    try:
         land_shp = shpreader.BasicReader(
-            current_app.config["SHAPE_FILE_DIR"] + "/cartopy_resources/ne_110m_diff.shp"
+            current_app.config["SHAPE_FILE_DIR"] + shp_file
         )
+    except shpreader.shapefile.ShapefileException:
+        land_shp = cfeature.NaturalEarthFeature("physical", "land", resolution)
```
htmlboss, line 36: nit: could you please add a basic print statement here so we can capture exception case in our rudimentary logs? Something like "using cartopy built-in image as fallback" or whatever.

htmlboss (approved):

Issue #971, opened 2022-02-22
Subset panel does not update when dataset is changed within the area window


JustinElms:
**Describe the bug**
The options in the subset panel are not updated when the dataset is changed from within the AreaWindow component. In the screenshots below I change datasets from RIOPS 3D Forecast to GIOPS 10 Day 3D forecast but the subset panel remains unchanged. The most notable issue is that these datasets are hourly and daily forecasts we should see the time pickers update to accommodate the time resolution of the second dataset. Clicking the 'save' button on the subset panel will download a netcdf file containing data from the first dataset.

**To Reproduce**
Create an area on the main map using any method and click plot if the area window hasn't appeared. Once the window opens select a different dataset from Main Map panel and click 'go'. You should find that the plot and other panels update while the subset panel does not. 

**Expected behavior**
The subset panel should update to display variables and time data from the new dataset. 

**Screenshots**
![subset_1](https://user-images.githubusercontent.com/79917349/155190140-a90754c7-5139-48f8-9c8c-aba49edfa028.png)
![subset_2](https://user-images.githubusercontent.com/79917349/155190132-a6c4ef07-a00b-4a80-9267-d94b6d689427.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 97.0.4692.99

PR #1030, opened 2022-06-29
Subset Panel Time Stamp Update Issue Resolved 

prodyut1978:
## Background
This PR is the continuation of the issue https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/971 - Subset panel does not update when dataset is changed within the area window. The previous PR # 971 only resolved the issue with subset variable update with the dataset changed. The present PR resolved the issue with timestamp update with the dataset changed.

The Time Stamp variable "Start Time (UTC)"  and "End Time (UTC)" in the subset panel not updating with the change of dataset and update of subset variables. The date and time was showing NAN and time selector calendar was remain unchanged (default Giops_Day time range). Please see below some of the screen shots of this issue:
Capture2
Capture3
Capture4

- Create a separate component of SubsetPanel.jsx and import it in the AreaWindow.jsx component
- The TimePicker.jsx component needed a componentDidUpdate method to query the timestamps for new datasets. 

## Screenshot(s)

Capture1

## Checks
- [y ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/kml
+++ b/kml
@@ -1,1 +1,1 @@
-/data/kml
+E:/data/kml
--- a/oceannavigator/frontend/src/components/AreaWindow.jsx
+++ b/oceannavigator/frontend/src/components/AreaWindow.jsx
@@ -6,8 +6,7 @@
 */
 
 import React from "react";
-import {Nav, NavItem, Panel, Row,  Col, Button, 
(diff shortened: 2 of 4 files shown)


prodyut1978:
Closed and made similar one

PR #1031, opened 2022-06-30
Quiver variable restriction on model class

prodyut1978:
## Background
The current Ocean navigator UI, the Quiver variable plot option is only working for "Mercator" model class type dataset. For example : 
"giops_day": {
"name": "01. GIOPS 10 day Forecast 3D - LatLon",
"model_class": "Mercator"
}
or 
"cmems-global-analysis-forecast-phy-001-024_hourly": {
        "name": "15. CMEMS Global Ocean PHYS Analysis (Hourly) 1/12 deg",
        "model_class": "Mercator
"}
There will be ongoing work in the future so that Quiver plot can work with the other model class Like "Nemo" or "fvcom" type dataset, example: 
"riops_fc_3dps": {
        "name": "06. RIOPS Forecast 3D - Polar Stereographic",
        "model_class": "Nemo"
}

For the end user, it is essential to hide and restrict the selection of the quiver variables for non  "Mercator" type model class to avoid 500 (INTERNAL SERVER ERROR).  The screen shots are below:
Capture6

Capture3

For the  "Mercator" type model class Quiver variable selection and plotting is working as before. The restriction is only implemented in the front end [DatasetSelector.jsx](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/compare/Quiver-variable-restriction-on-model-class?expand=1#diff-9a106321bb811a2c2f7cf5ca0934f1335595c9be3985610f880eec92e61a08fa) file and can be easily modified when other model class dataset (like "Nemo" or "fvcom") will be working. 

Below is the screen shot when user select "Nemo" model class type dataset "06. RIOPS Forecast 3D - Polar Stereographic".  The user not able to see any Quiver Variables to plot. 
Capture9

## Screenshot(s)

Capture

## Checks
- [Yes ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/DatasetSelector.jsx
+++ b/oceannavigator/frontend/src/components/DatasetSelector.jsx
@@ -36,6 +36,8 @@ const PARENT_ATTRIBUTES_TO_UPDATE = Object.freeze([
   "quiverVariable",
 ]);
 
+const MODEL_CLASSES_WITH_QUIVER = Object.freeze(["Mercator"]);
+
 class DatasetSelector extends React.Component {
   constructor(props) {
     super(props);
@@ -75,6 +77,7 @@ class DatasetSelector extends React.Component {
     });
 
     const quantum = currentDataset.quantum;
+    const model_class= currentDataset.model_class;
 
     GetVariablesPromise(newDataset).then(variableResult => {
       this.setState({ loadingPercent: 33 });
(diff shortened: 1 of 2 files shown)


oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -75,6 +77,7 @@ class DatasetSelector extends React.Component {
     });
 
     const quantum = currentDataset.quantum;
+    const model_class= currentDataset.model_class;
```
JustinElms, line 80:
```suggestion
    const model_class = currentDataset.model_class;
```

JustinElms (approved):

PR #1033, opened 2022-07-01
Subset Panel Time Stamp Update Issue Resolved

prodyut1978:
## Background
This PR is the continuation of the issue https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/issues/971 - Subset panel does not update when dataset is changed within the area window. The previous PR # 971 only resolved the issue with subset variable update with the dataset changed. The present PR resolved the issue with timestamp update with the dataset changed.

The Time Stamp variable "Start Time (UTC)" and "End Time (UTC)" in the subset panel not updating with the change of dataset and update of subset variables. The date and time was showing NAN and time selector calendar was remain unchanged (default Giops_Day time range). Please see below some of the screen shots of this issue:
image
image
image

Create a separate component of SubsetPanel.jsx and import it in the AreaWindow.jsx component. The TimePicker.jsx component needed a componentDidUpdate method to query the timestamps for new datasets.

## Screenshot(s)
image

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/SubsetPanel.jsx
```diff
@@ -0,0 +1,242 @@
...
+    </div>
+    );
+  }   
+}
+
+//***********************************************************************
+SubsetPanel.propTypes = {
+  dataset: PropTypes.object.isRequired,
+  area: PropTypes.array.isRequired,
+};
```
JustinElms, line 240: I think `id` and `key` should also be listed here

JustinElms:
Looks good overall, a couple suggestions:
- Check out the LGTM and CodeFactor errors and fix as much of them as you can. Most of these appear to be small things like long lines but there's a couple related to state handling and unused imports that should be addressed. 
- I noticed that there's some weird behavior with the timepicker. It looks like the End Time picker isn't being updated properly; when switching datasets the calendar suggests that there's data available every day up to the latest time stamp, selecting one of the dates outside of the range of the dataset causes the timepicker to display NaN for daily datasets or an empty hour selector for hourly ones. 
You might be able to fix that by modifying the `componentDidUpdate` method we added to the timepicker but a better option might be to hold off rendering the `SubsetPanel` component until the the `DatasetSelector` has loaded. This way you could remove the `componentDidUpdate` method from the timepicker altogether. To do this you will probably have to pass the `loading` state from `DatasetSelector` to `OceanNavigator` components to make it global then pass it back down as a prop but this could be useful for other components down the road. @htmlboss might have a better suggestion though. 
image
image

JustinElms:
Remove `const stringify = require("fast-stable-stringify");` on line 26 of AreaWindow.jsx to fix the unused import error.

JustinElms (approved):
Not sure what's going on with those CodeFactor but you clearly fixed the long lines it's still complaining about. The `DOM text reinterpreted as HTML` error might be something we should revisit down the road, otherwise this looks good!

PR #1047, opened 2022-07-21
Add Accordion Component for UI Organization

JustinElms:
## Background
Modern web applications make use of a collapsible accordion element to hide infrequently used options or elements. We have a number of places where the UI is fairly busy and using this types of component could be help improve the Navigator UX. The plot window settings panel is one such place. The accordion was added to hide the plot size and title options in these windows so that the user can focus on the plot data and more frequently used options.

Just used the `Panel` components to create an accordion component what can be inserted wherever needed. This can be reused in many places to it makes sense to put it into its own component. 

The `Panel` component expects an `onToggle` callback even though its not needed (and not shown in the [documentaion](https://react-bootstrap-v3.netlify.app/components/panel/#panels-collapsible)) otherwise it produces an error everytime its rendered. A workaround is to add a no-op function but it creates a LGTM error as seen bellow. See https://github.com/react-bootstrap/react-bootstrap/issues/3141 for more info.  

## Screenshot(s)
accord

## Checks
- [ ] I ran unit tests.


oceannavigator/frontend/src/components/lib/Accordion.jsx
```diff
@@ -0,0 +1,52 @@
...
+      panelIcon: "angle-down"
+    };
+  };
+
+  render() {
+    return (
+      <Panel 
+        id={this.props.id}
+        expanded={this.state.open}
+        onToggle={}
```
JustinElms, line 23: Necessary to suppress an error from React. See https://github.com/react-bootstrap/react-bootstrap/issues/3141.

dwayne-hart (approved):

PR #1048, opened 2022-07-25
Pathlib Implementation In Ocean Navigator UI - The Process Of FastAPI Migration

prodyut1978:
## Background
The present Ocean Navigator UI was implemented with OS library to handle all the basic file system operations. However it is currently recommended from the Python user community to step up to another library for most of the basic file system operations. That is the Pathlib, which is also a Python built-in library. It is more intuitive in terms of syntax, easier to use and has more features out of the box. The PathLib module offers classes representing filesystem paths with semantics appropriate for different operating systems. The details link of the PathLib can be below:
https://docs.python.org/3/library/pathlib.html.
There is also a link below where it is highly recommended to implement PathLib module over OS module. 
https://towardsdatascience.com/dont-use-python-os-library-any-more-when-pathlib-can-do-141fefb6bdb5
The Ocean Navigator developer team is currently in the process of FastAPI Migration from Flask. In this process we have planned to replace OS module and implement PathLib for all the file systems operations.

In this PR we have replaced all the OS related path with PathLib in the main UI. Below are the list of folders in the UI where all OS related file path were replaced:
- data
- oceannavigator
- plotting
- routes
- utils

The tests folder with all the Unit tests files OS modules will be replaced in different PR. 

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


routes/api_v1_0.py
```diff
@@ -1040,7 +1041,7 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
     """
     cache_dir = current_app.config["CACHE_DIR"]
     shape_file_dir = current_app.config["SHAPE_FILE_DIR"]
-    requestf = str(os.path.join(cache_dir, request.path[1:]))
+    requestf = str(Path(cache_dir).joinpath(request.path[1:]))
     basedir = requestf.rsplit("/", 1)[0]
```
JustinElms, line 1045:
``` suggestion 
    requestf = Path(cache_dir).joinpath(request.path[1:])
    basedir = requestf.parents[0]
```

routes/api_v1_0.py
```diff
@@ -1051,7 +1052,7 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
         return send_file(shape_file_dir + "/blank.mbt")
 
     # Send file if cached or select data in SQLite file
-    if os.path.isfile(requestf):
+    if Path(requestf).is_file():
```
JustinElms, line 1055:
``` suggestion
    if requestf.is_file():
```

routes/api_v1_0.py
```diff
@@ -1064,8 +1065,8 @@ def mbt(projection: str, tiletype: str, zoom: int, x: int, y: int):
         return send_file(shape_file_dir + "/blank.mbt")
 
     # Write tile to cache and send file
-    if not os.path.isdir(basedir):
-        os.makedirs(basedir)
+    if not Path(basedir).is_dir():
+        Path(basedir).mkdir(parents=True, exist_ok=True)
```
JustinElms, line 1069:
``` suggestion
    basedir.mkdir(parents=True, exist_ok=True)
````

File: routes/api_v1_0.py [hunk 4]
```diff
@@ -1498,9 +1499,9 @@ def _cache_and_send_img(bytesIOBuff: BytesIO, f: str):
     bytesIOBuff: BytesIO object containing image data
     f: filename of image to be cached
     """
-    p = os.path.dirname(f)
-    if not os.path.isdir(p):
-        os.makedirs(p)
+    p = PurePath(f).parent
+    if not Path(p).is_dir():
+        Path(p).mkdir(parents=True, exist_ok=True)
```
JustinElms, line 1504: It's better to use `Path()` instead of `PurePath()` in places like this. Anytime you need to do an IO operation like `mkdir` you can just create one `Path()` object and use it as needed. 

Something like this would probably work here:
``` suggestion 
    p = Path(f).parent
    if not p.is_dir():
        p.mkdir(parents=True, exist_ok=True)
```

oceannavigator/__init__.py
```diff
@@ -87,8 +88,8 @@ def create_app(testing: bool = False, dataset_config_path: str = ""):
     app.git_tag = subprocess_check_output("git", "describe", "--tags", "--abbrev=0")
 
     if app.config.get("WSGI_PROFILING"):
-        if not os.path.isdir("./profiler_results"):
-            os.mkdir("./profiler_results")
+        path = Path("./profiler_results")
+        Path(path).mkdir(parents=True, exist_ok=True)
```
JustinElms, line 92: You call `Path()` on path objects in a number of places. Once you create the path object you can use it without calling `Path()` again. 

``` suggestion
        path = Path("./profiler_results")
        path.mkdir(parents=True, exist_ok=True)
```

plotting/colormap.py
```diff
@@ -287,11 +287,11 @@ def find_colormap(name):
...
     "thermal": cmocean.cm.thermal,
     "neo_sst": mcolors.ListedColormap(
-        np.loadtxt(os.path.join(data_dir, "neo_sst.txt"))
+        np.loadtxt(Path(data_dir).joinpath("neo_sst.txt"))    
     ),
-    "BuYlRd": mcolors.ListedColormap(np.loadtxt(os.path.join(data_dir, "BuYlRd.txt"))),
+    "BuYlRd": mcolors.ListedColormap(np.loadtxt(Path(data_dir).joinpath("BuYlRd.txt"))),
     "temperature": mcolors.ListedColormap(
-        np.loadtxt(os.path.join(data_dir, "temperature.txt"))
+        np.loadtxt(Path(data_dir).joinpath("temperature.txt"))
```
JustinElms, line 294: Just use `data_dir.joinpath()` here instead of calling `Path()` on data_dir again.

JustinElms:
I think this is just about done. You just need to fix the path calls in that last area. Also, if you haven't run Black on the files you modified already I'd do that. Don't worry about using it on `api_v1_0.py` though, there's too much code in there that might break and we're about to rewrite it for the FastAPI migration anyway.

plotting/colormap.py
```diff
@@ -287,11 +287,11 @@ def find_colormap(name):
     "potential sub surface channel": mcolors.ListedColormap(["#ecf0f1", "#f57732"]),
     "thermal": cmocean.cm.thermal,
     "neo_sst": mcolors.ListedColormap(
-        np.loadtxt(os.path.join(data_dir, "neo_sst.txt"))
+        np.loadtxt(data_dir.joinpath("neo_sst.txt"))
```
JustinElms, line 290:
``` suggestion
        np.loadtxt(data_dir.joinpath("neo_sst.txt"))
````

File: routes/api_v1_0.py [hunk 8]
```diff
@@ -1476,13 +1474,13 @@ def _is_cache_valid(dataset: str, f: str) -> bool:
     """
 
     config = DatasetConfig(dataset)
-    if os.path.isfile(f):
+    if Path(f).is_file():
         cache_time = config.cache
         if cache_time is not None:
-            modtime = datetime.datetime.fromtimestamp(os.path.getmtime(f))
+            modtime = datetime.datetime.fromtimestamp(f.stat().st_mtime)
```
JustinElms, line 1480: Is `f` a Path object here? I ask because you use `Path(f)` on 1477 and 1483. If not, create one Path object from `f` above and use it accordingly.

prodyut1978: Hi Justin, f is a poxis path, called from tiles route. I changed it accordingly.
Path(f).is_file(): to f.is_file():

and also f.unlink() in line 1483

JustinElms (approved):

PR #1050, opened 2022-08-01
API Script R Subset Implementation- The Process Of FastAPI Migration

prodyut1978:
## Background
The API R Script for downloading subset variables time series .nc data  was not implemented in the present Ocean navigator UI. This PR added the rSUBSETtemplate.txt for R template to download subset variables .nc data.

The client requested to implement the API R script fo subset variables. 

## Screenshot(s)
Capture

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/plotting/scriptGenerator.py
+++ b/plotting/scriptGenerator.py
@@ -56,8 +56,6 @@ def generateR(url, scriptType: str) -> BytesIO:
 
         if scriptType == "PLOT":
             template = template.format(q=query, var=var)
-        elif scriptType == "SUBSET":
-            template = template.format(q=query, init='{', end='}')
         else:
             template = template.format(q=query)
 
--- a/plotting/scriptGenerator.py
+++ b/plotting/scriptGenerator.py
@@ -56,6 +56,8 @@ def generateR(url, scriptType: str) -> BytesIO:
 
         if scriptType == "PLOT":
             template = template.format(q=query, var=var)
+        elif scriptType == "SUBSET":
+            template = template.format(q=query, init='{', end='}')
         else:
             template = template.format(q=query)
 
--- a/plotting/templates/rSUBSETtemplate.txt
+++ b/plotting/templates/rSUBSETtemplate.txt
@@ -10,7 +10,7 @@ url <- base_url
 
 for (key in names(queryObj))
 {{
- url <- paste(url,"&", key, "=", queryObj[key], sep="")
+   url <- paste(url, "&", key, "=", queryObj[key], sep="")
 }}
 
 url  <- gsub(",[\r\n]", "", url)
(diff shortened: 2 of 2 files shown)


plotting/templates/rSUBSETtemplate.txt
```diff
@@ -0,0 +1,17 @@
+library(jsonlite)
+
+query <- '{q}'
+queryObj <- fromJSON(query)
+fname <- paste("script_output",".nc")
```
JustinElms, line 5: `paste` concatenates strings. Since we're not adding anything to the output filename you can just set
`fname <- "script_output.nc"`

plotting/scriptGenerator.py
```diff
@@ -56,6 +56,8 @@ def generateR(url, scriptType: str) -> BytesIO:
 
         if scriptType == "PLOT":
             template = template.format(q=query, var=var)
+        elif scriptType == "SUBSET":
+            template = template.format(q=query, init='{', end='}')
```
JustinElms, line 60: You shouldn't need to specify `init='{', end='}' here. If you're getting an error is probably because of the `str.format()` function is confused by the for loop curly braces in the R template. I think the format will work if you double the braces. Something like this in the R template should work:
```
for (condition) {{
    do thing
}}
```

JustinElms:
A couple small changes needed that I detailed above. The resulting R script is also a little congested. Adding some line breaks and indents where appropriate can go a long way in making it more readable.

plotting/scriptGenerator.py
JustinElms: Also, if you remove the `init='{', end='}'` you wont need the `elif scriptType == "SUBSET":` condition.

plotting/templates/rSUBSETtemplate.txt
```diff
@@ -0,0 +1,20 @@
+library(jsonlite)
+
+query <- '{q}'
+
+queryObj <- fromJSON(query)
+fname <- "script_output",".nc"
```
JustinElms, line 6:
``` suggestion
fname <- "script_output.nc"
```

plotting/templates/rSUBSETtemplate.txt
```diff
@@ -0,0 +1,20 @@
...
+
+base_url <- "http://navigator.oceansdata.ca/api/v1.0/subset/?"
+url <- base_url
+
+for (key in names(queryObj))
+{{
+ url <- paste(url,"&", key, "=", queryObj[key], sep="")
```
JustinElms, line 13:
``` suggestion
   url <- paste(url, "&", key, "=", queryObj[key], sep="")
```

JustinElms (approved):

Issue #1055, opened 2022-08-15
Additional contour plot not working in the area plot window


prodyut1978:
**Describe the bug**
The addition of contour plot in the selected area plot window is showing backend error.

**To Reproduce**

![ContourPlotissue](https://user-images.githubusercontent.com/80789651/184716842-7e08f1b9-7301-42e7-84c2-05c4dd05bd91.gif)

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #1056, opened 2022-08-15
Additional contour plot issue resolved

prodyut1978:
## Background
The Additional Contours plot in the area plot window was showing error in the backend. 
Capture
Capture9
This PR resolved the below issue:
Additional contour plot not working in the area plot window #1055

## Screenshot(s)

Capture
Capture1

##
 Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/plotting/map.py
+++ b/plotting/map.py
@@ -106,7 +106,7 @@ class MapPlotter(Plotter):
             and self.quiver["variable"]
             and self.quiver["variable"] != "none"
         )
- 
+
     def __load_contour(self) -> bool:
         return (
             self.contour is not None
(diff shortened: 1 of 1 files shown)


plotting/map.py
```diff
@@ -390,6 +401,7 @@ def load_data(self):
                 contour_unit = vc.unit
                 contour_name = vc.name
                 contour_data.append(d)
+                self.contour_data = contour_data
```
JustinElms, line 404: We set `self.contour_data = contour_data` here and outside of the _if_ block on line 408. We probably don't need both so I think we should just keep the one on 408 so that `self.contour_data` will be empty when there's no contours.

JustinElms:
On lines 386-390:

You can replace 
``` 
            if (
                self.contour is not None
                and self.contour["variable"] != ""
                and self.contour["variable"] != "none"
            ):
```
with 
```
if self.__load_contour():
```
since its the same check.

JustinElms:
Looks good overall! Just need to fix the two small things I mentioned above. Also, make sure you format the map.py file with Black if you haven't already.

JustinElms (approved):

PR #1066, opened 2022-08-31
Allow entering/importing point lattitude- longitudes values outside [-90, 90] and [-180, 180] range

prodyut1978:
## Background
**Describe the bug**
Using longitude values outside of the -180 to 180 degree range in the the Enter Point/Line/Area windows results in some odd behavior from the Navigator. On the front end side the map centers on the resulting point and zooms in to the maximum level. Meanwhile, the backend doesn't seem to be able to produce the requested data resulting in an empty plot. Using a longitude value within the [-180,180] degree range (ie -155 degrees instead of 205) works as expected.

Because this bug affects both front and back ends of the Navigator we probably just need to cast the longitude value to the [-180,180] range at the text input level.

**To Reproduce**
Open any of the Enter Point/Line/Area windows and enter a longitude coordinate outside of the range. you should see the behavior described above.

**Expected behavior**
The Navigator should respond the same way to equivalent longitude values.

UI cleanup in the process of fast API migration

Latitude ranges between -90 and 90 degrees, inclusive. Values above or below this range will be clamped to the range [-90, 90]. This means that if the value specified is less than -90, it will be set to -90. And if the value is greater than 90, it will be set to 90.
 Longitude ranges between -180 and 180 degrees, inclusive. Values above or below this range will be wrapped so that they fall within the range. For example, a value of -190 will be converted to 170. A value of 190 will be converted to -170. This reflects the fact that longitudes wrap around the globe
## Screenshot(s)

Capture1
Capture
Capture2

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/MapToolbar.jsx
```diff
@@ -312,11 +312,26 @@ class MapToolbar extends React.Component {
...
+    let latitude = this.state.coordinate[0]
+    let longitude = this.state.coordinate[1]
+    //Latitude ranges between -90 and 90 degrees, inclusive. 
+    //Values above or below this range will be clamped to the range [-90, 90]. 
+    //This means that if the value specified is less than -90, it will be set to -90. 
+    //And if the value is greater than 90, it will be set to 90.
+    latitude = (latitude >90) ? 90: latitude;
+    latitude = (latitude <-90) ? -90: latitude;
+    //Longitude ranges between -180 and 180 degrees, inclusive. 
+    //Values above or below this range will be wrapped so that they fall within the range.
```
JustinElms, line 324: Rather than forcing the latitude to be +/- 90 when outside the range it might be better have the value wrap around the globe. For example a user input of 100 degrees latitude would be converted to 80 degrees, 200 to -20 (very unlikely but should be considered), etc.

oceannavigator/frontend/src/components/MapToolbar.jsx
```diff
@@ -312,11 +312,26 @@ class MapToolbar extends React.Component {
...
+    //For example, a value of -190 will be converted to 170. A value of 190 will be converted to -170. 
+    //This reflects the fact that longitudes wrap around the globe.//
+    longitude = (longitude >180) ? longitude - 360: longitude;
+    longitude = (longitude <-180) ? 360 + longitude: longitude;
+    let appliedPointCords = [latitude, longitude]
```
JustinElms, line 329: It might be better to use the % operator here so that you can handle unexpectedly large inputs. You might be able to combine these statements into one line with it too.

oceannavigator/frontend/src/components/lib/CheckBox.jsx
```diff
@@ -31,8 +31,8 @@ export default class CheckBox extends React.PureComponent {
 //***********************************************************************
 CheckBox.propTypes = {
   title: PropTypes.string.isRequired,
-  checked: PropTypes.bool.isRequired,
+  checked: PropTypes.bool,
```
JustinElms, line 34: Was this causing an error? I would imagine it's necessary for the `CheckBox` component.

prodyut1978: Yes It causing an warning, because not all calling Checkbox component uses this checked option. Like below on Mapinput.jsx file >>>>>
                  <CheckBox
                    id='syncRanges'
                    onUpdate={this.props.changeHandler}
                    title={_("Sync Variable Ranges")}
                    style={{display: this.props.dataset_compare ? "block" : "none"}}
                  />

oceannavigator/frontend/src/components/MapToolbar.jsx
prodyut1978: Actually I was using the concept from the below link:
https://developers.google.com/maps/documentation/javascript/reference/coordinates

PR #1100, opened 2023-10-13
Fix Linear Threshold feature

JustinElms:
## Background
It looks like the Linear Threshold feature was broken during the UI updated (v9.0). The issues appears to lie within the _NumberBox_ component where two problems occur: the underlying `<input>`'s event was passed to the `this.changed` event handler instead of just the event value and there was no validation in place to ensure that the received value was a number. 

To restore functionality we now just pass the event's value to the input change handler and added validation to ensure that this is a number. 

Following best practices for a numeric input in React.

Fixed some code formatting in _LineWindow_ and _NumberBox_ components.

## Checks
- [ ] I ran unit tests.

### DIFF
--- a/oceannavigator/frontend/src/components/NumberBox.jsx
+++ b/oceannavigator/frontend/src/components/NumberBox.jsx
@@ -1,6 +1,5 @@
 import React from "react";
 import { Modal, Button } from "react-bootstrap";
-// import NumericInput from "react-numeric-input";
 import Icon from "./lib/Icon.jsx";
 import PropTypes from "prop-types";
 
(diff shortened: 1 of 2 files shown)


oceannavigator/frontend/src/components/NumberBox.jsx
```diff
@@ -28,9 +27,12 @@ class NumberBox extends React.Component {
   changed(num, str) {
     clearTimeout(this.timeout);
 
-    this.setState({
-      value: num,
-    });
+    num = Number(num)
```
Neha-Thakare1995, line 30: We can also try required attribute on Input Tag.

PR #1104, opened 2023-11-01
Add Hide Data Layer button to MapInputs panel

JustinElms:
## Background
We received a request to provide a way to hide the data layer so that users can view the bathymetry below. A practical solution was to add a _Hide Data Layer_ button to the MapInputs panel which toggles the data layer's visibility in the MainMap component via the `mapSettings` prop.

This is a intuitive way to allow users to hide the data layer and was straightforward to implement. 

The button currently only toggles the data layer. Other layers, such as drawing layers, quivers, and observations, can easily be added to this functionality if requested.  

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.


NehaThakare (approved):
Looks good to me

PR #1110, opened 2024-01-10
Optimize Observation Database Queries

Neha-Thakare1995:
## Background
- Updated the observation queries to optimize the database. Change the get_stations() to change the query, to reduce the time take to get the response from the query. 
- Added the "idx_stations_time" index to the station table.
- Changed the joint functionality to reduce the response time of the query.  

- The query use to take almost nearly 10 to 12 minutes to get the response for the observation query to get the stations.
- This has been now reduced to 1 min to get the response of the get_stations() with the updated query.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


data/observational/queries.py
```diff
@@ -372,14 +376,31 @@ def get_stations(
...
+            meta_value=meta_value,
+        )
+        .join(Platform)
+        .all()
+    )
 
 
 def __get_bounding_latlon(lat, lon, distance):
-    # angular distance in radians on a great circle
+    # angular di      stance in radians on a great circle
```
JustinElms, line 403:
```suggestion
# angular distance in radians on a great circle
```

JustinElms (commented):
Great work Neha! Just need to remove the `querys.ipynb` file you used for testing then this PR will be good to go!

Neha-Thakare1995 (commented):
Please have a look on the made changes.
Thank You!

JustinElms (approved):

PR #1113, opened 2024-01-23
Seawater to gsw

bbbrandyn:
## Background
seawater package is deprecated. Replaced with gsw

gsw uses the TEOS-10 instead of the EOS-80 used by 'seawater' which is now considered obsolete

tempgradient() function now returns result in K/Pa rather than C/db. This change needs to be reflected in configs branch

## Checks
- [X] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/config/conda/environment.yml
+++ b/config/conda/environment.yml
@@ -97,6 +97,7 @@ dependencies:
   - greenlet=3.0.2=py312h30efb56_0
   - gst-plugins-base=1.22.7=h8e1006c_1
   - gstreamer=1.22.7=h98fc4e7_1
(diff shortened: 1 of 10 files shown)


JustinElms (commented):
It looks like seawater is still listed in your environment yml and spec files. Are you able to remove it from the environment and regenerate those files or is a dependency for another package?

JustinElms:
See [#63](https://github.com/DFO-Ocean-Navigator/configurations/pull/63) for change of _tempgradient_ units in config files.

JustinElms (approved):
Great work Brandyn!

PR #1117, opened 2024-02-05
Seawater to gsw

bbbrandyn:
the seawater package used a positive depth value, whereas gsw calculates depth as negative. Since this wasn't noticed before the previous PR, any variable using depth was giving unexpected results. fixed the code that involves depth.

## Checks
- [X] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @dwayne-hart @JustinElms @VanessaSutton-Pande
+* @JustinElms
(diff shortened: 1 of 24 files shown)


JustinElms:
@bbbrandyn Looks like there are some merge conflicts here that need to be resolved. Otherwise the changes look good!

JustinElms:
These changes look good but you'll have to resolve the merge conflicts listed below. This PR's documentation could use some context as well. You could give some more detail about the issue caused by the previous PR and include a screenshot of the broken behavior.

JustinElms (approved):

Issue #1105, opened 2023-11-08
Changing dataset in area window with subset variable causes crash


JustinElms:
**Describe the bug**
Found a n odd scenario that causes a crash when I was testing the subset feature this week. The Navigator crashes if a user changes the selected dataset in the area window after they have selected a variable in the subset panel. 

**To Reproduce**
Open an area plot in the Navigator then click on one or more variables in the subset pane variable selector. Once selected change the current dataset. At this point the Navigator will crash resulting in a white screen. 

**Expected behavior**
The dataset should change and the variable selector should update with the new dataset's variables.

PR #1119, opened 2024-02-08
area plot crash fix

bbbrandyn:
## Background
Issue #1105 . While dataset is loading the values are set to null which causes a crash on update.

I edited the ComboBox.js file to skip the update if the values from 'dataset0' are null. This will skip the updates that happen while the dataset is loading. 

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


JustinElms:
Code changes look good but documentation needs to be expanded. You can link to the issue by mentioning it in the text above. You should also assign yourself and add the relevant labels to the PR.

JustinElms (approved):

PR #1121, opened 2024-02-22
Fix EnterCoordsWindow crash

JustinElms:
## Background
The coordinate fields in the `EnterCoordsWindow` caused the Navigator to crash when the values were deleted. To resolved we added some additional input validation before updating the `vectorCoords` prop to avoid the crash.

This approach ensures that valid values are passed to the `vectorCoords` prop which is mapped to the input fields. 

## Checks
- [ ] I ran unit tests.


Neha-Thakare1995 (approved):
Changed code version says the better version of the previous one. By wrapping props.action() in an arrow function inside setTimeout, ensures that props.action() is called after the 1000ms delay.

Where as previous implementation was directly passing the result of props.action to setTimeout, which was getting executed immediately rather than after the specified timeout.

Seems good to me.

PR #1122, opened 2024-02-27
Fix daily calendar selector for CMEMS datasets

JustinElms:
## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


bbbrandyn (approved):
nice fix, looks good

PR #1125, opened 2024-02-28
Limit points in timeseries plot

JustinElms:
## Background
VM/Timeseries plot requests covering more than ~50 timestamps have historically taken longer than the 5 minute response time limit allotted to Navigator queries. In the past we have limited the timepicker components so that users cannot select timespans greater than 50 timestamps but this was dropped when the UI was refactored last year. Instead of taking a similar approach I opted to subsample the timestamps so than users can request timeseries plots over any length of time but the result plot will only have a maximum of 50 points.    

Unlike the previous approach described above, this allows users to plot any length of timeseries.

## Checks
- [X] I ran unit tests.

### DIFF
--- a/data/netcdf_data.py
+++ b/data/netcdf_data.py
@@ -996,6 +996,10 @@ class NetCDFData(Data):
                 raise RuntimeError("Error finding timestamp(s) in database.")
 
             file_list = db.get_netcdf_files(timestamp, variables_to_load)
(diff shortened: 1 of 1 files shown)


Neha-Thakare1995 (approved):
Seems good to me. It limits the NetCDF files instead of loading for long time.

PR #1128, opened 2024-03-20
Cmems data_importers scripts

Neha-Thakare1995:
## Background

This pull request addresses the need to import data from NetCDF files from cmems into a database using Python. The code leverages the xarray library to handle NetCDF files and SQLAlchemy for database operations. The script aims to provide a flexible and efficient solution for importing observational data into a database.

1. Database operations are optimized by using SQLAlchemy's session management. Sessions are committed after each step to minimize the risk of data loss in case of failures.
2. To import data from NetCDF files from cmems into a database using Python.

1. The script iterates over each NetCDF file, extracting relevant data variables defined in the VARIABLE_MAPPING. It then iterates over the rows of the DataFrame to create corresponding database records for stations, samples, data types, and platforms.
3. Users can specify the database URI as an input parameter, allowing them to connect to different databases without modifying the code.
4. The defopt library is utilized to automatically generate a CLI for the script, simplifying its invocation and usage. Users can run the script from the command line with minimal effort.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


scripts/data_importers/cmems_drifter.py
```diff
@@ -0,0 +1,106 @@
...
+                            unit=unit,
+                        )
+                        session.add(data_type)
+                        session.commit()
+
+                    # Create Platform object
+                    platform = Platform(
+                        type=Platform.Type.drifter,
+                        # attributes={key: str(ds.attrs[key]) for key in ds.attrs},
+                    )
```
JustinElms, line 71:
``` suggestion
                    platform = Platform(type=Platform.Type.drifter)
```

Neha-Thakare1995:
@JustinElms The script has been as per the suggestion

JustinElms (approved):

JustinElms:
Great work Neha!

PR #1134, opened 2024-03-25
include quiver when updating map from mountedDataset

bbbrandyn:
## Background
on #1116 I overlooked including quiver when updating the map from mountedDataset. Added now so that "Get Link" will include state of quiverVariable and quiverDensity

Copied the process used for the other variables in moountedDataset

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


JustinElms (approved):
Awesome, thanks for taking care of this so quickly!

Issue #1135, opened 2024-03-28
Add scale selector to area plot windows


JustinElms:
**Priority (low, medium, high)**
Priority: high

**Is your feature request related to a problem? Please describe.**
Received a request to add a scale selector to the area plot window.

**Describe the solution you'd like**
A specific implementation has not been requested by the axis range selector added to the profiler and virtual mooring plotters might be an effective solution.

**Additional context**
User specifically asked to display the area plot with the same scale as the global map. The range selector mentioned above would cover that but may also need a "sync to global map" type button to cover the case where the main map scale has changed.

PR #1144, opened 2024-04-09
Add scale selector to AreaWindow

bbbrandyn:
## Background
#1135 

I decided to use the existing ColormapRange component to avoid redundancy. 

I'm not sure why the check box isn't in line with the other ones in the area settings card.

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/AreaWindow.jsx
+++ b/oceannavigator/frontend/src/components/AreaWindow.jsx
@@ -14,6 +14,7 @@ import SubsetPanel from "./SubsetPanel.jsx";
 import PropTypes from "prop-types";
 
 import { withTranslation } from "react-i18next";
+import { formToJSON } from "axios";
 
 class AreaWindow extends React.Component {
   constructor(props) {
@@ -24,8 +25,8 @@ class AreaWindow extends React.Component {
 
     this.state = {
       currentTab: 1, // Currently selected tab
-      scale: props.dataset_0.variable_scale,
-      scale_1: props.dataset_1.variable_scale_1,
+      scale: props.dataset_0.variable_scale[0] + "," +props.dataset_0.variable_scale[1] + ",auto",
+      scale_1: props.dataset_1.scale_1 + ",auto",
       scale_diff: "-10,10,auto",
       leftColormap: "default",
       rightColormap: "default",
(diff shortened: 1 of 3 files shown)


bbbrandyn:
> Looks good so far! How does it behave when the compare dataset option is checked?

It seems to work fine however I am not too familiar with expected behaviour of the compare datasets functionality

JustinElms:
It looks like the scale defaulted to -10, 10 when compare datasets was checked. Is it possible to update the scale here to do the same thing? Perhaps it should default to "Auto" when the compare datasets checkbox is first checked and if the user then wants to adjust the scale the default values can be -10,10.If that makes sense?

bbbrandyn:
sure I'll give that a try

bbbrandyn:
> It looks like the scale defaulted to -10, 10 when compare datasets was checked. Is it possible to update the scale here to do the same thing? Perhaps it should default to "Auto" when the compare datasets checkbox is first checked and if the user then wants to adjust the scale the default values can be -10,10.If that makes sense?

I'm not sure how to edit the state of the ColormapRange component from another function. Maybe the approach should be to hide the current ColormapRange component and show a separate one when Compare Datasets is selected?

JustinElms:
Maybe you can add an update function in AreaWindow that gets called when the box gets checked/unchecked and sets this.state.scale appropriately. Something like this might do it: 

```
// new update function could look something like this (add before the render/return):
compareChanged(checked) {

  if (checked) {
    // set the scale state to the compare range (-10,10)
  } else {
    // set the scale state to the variable range (-5,30, or whatever the default is for the variable)
  }

  this.props.setCompareDatasets(checked)
}

...

 <CheckBox
            id="dataset_compare"
            key="dataset_compare"
            checked={this.props.dataset_compare} 
            //onUpdate={(_, checked) => this.props.setCompareDatasets(checked)} 
            onUpdate={(_, checked) => compareChanged(checked)}
            title={_("Compare Datasets")}
          />
```

bbbrandyn:
I tried that, it did set the scale however it doesn't change the values in the input fields of the ColormapRange component

bbbrandyn:
I also think scale_1 and scale_diff are unnecessary now. Do you know why the auto checkbox is not in line?

JustinElms:
Its probably set in `Ocean-Data-Map-Project/oceannavigator/frontend/src/stylesheets/components/_ColormapRange.scss`

bbbrandyn:
added css changes

JustinElms (approved):

PR #1145, opened 2024-04-15
pyinstrument middleware added

bbbrandyn:
## Background
Allows for profiling api requests for a breakdown of time resources used

https://pyinstrument.readthedocs.io/en/latest/guide.html#profile-a-web-request-in-fastapi

To invoke, make any request with the GET parameter 'profile=1' and it will print the HTML result from pyinstrument.

example: http://localhost:8443/api/v2.0/tiles/giops_day/votemper/2337379200/0/4/9/3?projection=EPSG:3857&scale=-5,30&interp=gaussian&radius=25&neighbours=10&profile=1

## Screenshot(s)
unnamed

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/__init__.py
+++ b/oceannavigator/__init__.py
@@ -5,10 +5,11 @@ import dask
 import sentry_sdk
 from fastapi import FastAPI, Request
 from fastapi.middleware.gzip import GZipMiddleware
-from fastapi.responses import JSONResponse
+from fastapi.responses import JSONResponse, HTMLResponse
 from fastapi.routing import Mount
 from fastapi.staticfiles import StaticFiles
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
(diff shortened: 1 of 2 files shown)


bbbrandyn:
tests failing on importing the pyinstrument package. Does it need to be added to yaml?

JustinElms:
Oh yeah that'll do it. You can dump you conda environment into the yml and spec files with these commands:
`conda env export > config/conda/environment.yml`
`conda list --explicit > config/conda/navigator-spec-file.txt`

JustinElms:
Is it possible to write the output of pyinstrument to a log file rather than return it to the browser? This would allow us to profile the application while using the UI and make it easier to gather statistics.

JustinElms (approved):

PR #1146, opened 2024-04-23
Transect plot y-axis scale behaviour update

bbbrandyn:
## Background
Currently, a transect plot would have an exponential y-axis with an adjustable linear threshold by default. These changes change the default to a linear y-axis with a checkbox which enables the original (exponential with linear threshold) behaviour.

I retooled the DepthLimit component into a more general TransectLimiter component so that it could be used for both it's original purpose as well as the exponential toggle.

## Screenshot(s)
image

## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/AreaWindow.jsx
+++ b/oceannavigator/frontend/src/components/AreaWindow.jsx
@@ -49,7 +49,7 @@ class AreaWindow extends React.Component {
       syncLocalToGlobalState: false,
       showarea: true,
       surfacevariable: "none",
-      linearthresh: 200,
+      linearthresh: 0,
       bathymetry: true, // Show bathymetry on map
       plotTitle: undefined,
       quiver: {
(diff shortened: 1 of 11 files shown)


oceannavigator/frontend/src/components/AreaWindow.jsx
```diff
@@ -49,7 +49,7 @@ class AreaWindow extends React.Component {
       syncLocalToGlobalState: false,
       showarea: true,
       surfacevariable: "none",
-      linearthresh: 200,
+      linearthresh: 0,
```
JustinElms, line 52: Is `linearthresh` use in this component? If not we can just remove it

oceannavigator/translations/fr.json
bbbrandyn, line 1: Added this one. What is the purpose of the 'en.json'?

bbbrandyn:
added the french translation. Do I need to add something to the en.json as well?

JustinElms:
Yeah, technically it'll fall back on the provided text but en.json should get updated for good practice

JustinElms (approved):

Issue #1150, opened 2024-05-02
Add NetCDF sub-setting to  point plotter


JustinElms:
**Priority (low, medium, high)**
Priority: High

**Is your feature request related to a problem? Please describe.**
Received a request to expand NetCDF output to points.  

**Describe the solution you'd like**
We need some way to enable this feature. The point windows are already pretty busy so adding a subset panel similar to the one in the area window might not be the best approach. A better option might be to add a NetCDF option to the _Save Image_ button below the plot.

JustinElms:
Completed in #1152

PR #1152, opened 2024-06-10
Add NetCDF option to Profile and Virtual Morning Save Image menu

JustinElms:
## Background
We received a request to add a NetCDF file output option to the point plotters (see issue #1150). A _NetCDF_ button was added to the save image menu and related functionality added to the profiler and timeseries plotters. This feature should produce NetCDF files for any profile or VM plot regardless of the number of variables, points, and depths present in the current plot and the output will always contain the standard time, depth, lat, lon dimensions. 

Rather than recreate the subset window from the area plotter we decided that this should be a quick to use feature that allows users to quickly download the data visible in the current plot. 

## Screenshot(s)
image

## Checks
- [X] I ran unit tests.


bbbrandyn (approved):
Looked over code and tested, looks good

Issue #1140, opened 2024-04-05
Extract profile from transect plot


JustinElms:
**Priority (low, medium, high)**
Priority: medium

**Is your feature request related to a problem? Please describe.**
Users have requested the ability extract a vertical profile plot from a transect plot and display it in the same window. At this point the only way to achieve this is to add a point that intersects the transect and plot in a separate window which is cumbersome for users.

**Describe the solution you'd like**
Add UI controls to the Transect plot window that allows users to plot a profile by select a distance along the transect. The profile plot should be displayed in the same window. The control can either be a text box, a slider, or a combination and should be optional i.e. not a default behavior. The resulting profile plot should be queried separately so logic will have to be added to determine the lat/lon coordinates of the profile.

PR #1157, opened 2024-06-18
Transect profile

bbbrandyn:
## Background
#1140 

Can only specify a point between the first two points of the transect.

## Screenshot(s)
image

## Checks
- [X] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/MainMap.jsx
+++ b/oceannavigator/frontend/src/components/MainMap.jsx
@@ -1438,8 +1438,8 @@ const MainMap = forwardRef((props, ref) => {
   const getLineDistance = (line) => {
     var dist = 0;
     for (let i = 1; i < line.length; i++) {
-      let start = line[i - 1].reverse()
-      let end = line[i].reverse()
+      let start = [line[i - 1][1], line[i - 1][0]]
+      let end = [line[i][1], line[i][0]]
       dist += getDistance(start, end);
     }
 
--- a/oceannavigator/frontend/src/components/MainMap.jsx
+++ b/oceannavigator/frontend/src/components/MainMap.jsx
@@ -1436,7 +1436,10 @@ const MainMap = forwardRef((props, ref) => {
   };
 
   const getLineDistance = (line) => {
-    var dist = getDistance(line[0],line[1]);
+    var dist = 0;
+    for (let i = 1; i < line.length; i++) {
+      dist += getDistance(line[i-1],line[i]);
+    }
       
     return dist;
   }
--- a/oceannavigator/frontend/src/components/MainMap.jsx
+++ b/oceannavigator/frontend/src/components/MainMap.jsx
@@ -1438,9 +1438,11 @@ const MainMap = forwardRef((props, ref) => {
   const getLineDistance = (line) => {
     var dist = 0;
     for (let i = 1; i < line.length; i++) {
-      dist += getDistance(line[i-1],line[i]);
+      let start = line[i - 1].reverse()
+      let end = line[i].reverse()
+      dist += getDistance(start, end);
     }
-      
+
     return dist;
   }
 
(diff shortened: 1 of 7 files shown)


JustinElms:
Looks great so far! I think there's a couple small things we can do to help end users:
- Rename the checkbox to _Extract Profile Plot_ so its more consistent with the other ONAV plotters
- Add the point coordinates for the profile to the plot title so that users can see exactly where the profile is located
- The profile y-axis ticks are right next to color bar's units label so it looks like the profile plot's vertical axis label is the variable. That might be confusing for users so maybe we can remove the vertical ticks?
- Add the variable unit to the profile plot's horizonal axis label

After doing some digging this afternoon it sounds like OpenLayers includes some functionality to calculate distance between two points. We can probably make use of that on the front end to give users more control over the profile location. This super old thread has some examples that might be helpful:
https://stackoverflow.com/questions/10109620/openlayers-how-to-calculate-distance-between-two-points
but there might be a more modern way to do it in the OpenLayers docs: 
https://openlayers.org/

bbbrandyn:
Totally agree. I have pushed those readability changes.

I'll take a look at OpenLayers and see if we can have a more elegant solution for selecting the profile location. Also we may need to discuss how to handle transects with 3+ points as well as 'compare datasets' cases

JustinElms:
I'm not sure if you've tackled the multi-segment distance issue yet but I noticed this when I tried a line with three points:
image

bbbrandyn:
still looking into the multiple point functionality

bbbrandyn:
Stumped as to why OpenLayers is giving a different distance value than the plot. I believe they are both using the same EPSG coordinates system.

JustinElms:
> Stumped as to why OpenLayers is giving a different distance value than the plot. I believe they are both using the same EPSG coordinates system.

It's really unusual, I would have expected to get the same value too. I'll take a closer look at it tomorrow morning.

oceannavigator/frontend/src/components/MainMap.jsx
```diff
@@ -1433,6 +1435,15 @@ const MainMap = forwardRef((props, ref) => {
...
   };
 
+  const getLineDistance = (line) => {
+    var dist = 0;
+    for (let i = 1; i < line.length; i++) {
+      dist += getDistance(line[i-1],line[i]);
+    }
+      
+    return dist;
+  }
```
JustinElms, line 1445: You're getting a different distance here because your lat and lon values are in the wrong order. You could do something like this to swap them in the array and get the correct distance value:
```suggestion
  const getLineDistance = (line) => {
    var dist = 0;
    for (let i = 1; i < line.length; i++) {
      let start = line[i - 1].reverse()
      let end = line[i].reverse()
      dist += getDistance(start, end);
    }

    return dist;
  }
````

bbbrandyn: Huh, that was one of the first things I tried. I must have messed that up somehow 🤔

bbbrandyn:
pulled your changes and still have the issue. I noticed the original plot changes sizes when adding the profile plot. maybe it's an issue with the grid stuff?

JustinElms:
I was just doing some testing and found a couple small things that could be improved:
- Add a timeout to the distance slider to prevent it from sending multiple plot requests while the user is adjusting the value. This seems to happen when you drag the slider thumb rather than click on the distance. 
- Change the variable name on the profile x axis to the proper variable name. You can use the dataset_config.variable object to grab it.
- Double check that the vertical axis for the profile and transect are the same. Seems like the profile data extends deeper than the available data in the screenshot below:
image

bbbrandyn:
The current behaviour is that the plot will update when the user hasn't moved the slider for at least 1 second. What do you suggest? As for that plot, checking the profile location in the title it seems that it is extracting a profile from past the second point of the plot right?

bbbrandyn:
because of the way the line distance calculation is handled on the backend, it looks like it would be twice as slow if I passed the distance as a percentage from the frontend and recalculated it. So I think the current behaviour where there may be a slight discrepancy between frontend and backend distances would be preferable to that. What do you think?

JustinElms:
If the data in the plot is accurate then I don't think it will be a problem as is.

JustinElms (approved):

PR #1186, opened 2025-05-22
Make prints in scripts python3 compatible

DavidVadnais:
## Background
I wanted to look at some of the scripts but I did not want to install python2.  Then I ran `black .` and that changed other files.

This was the least invasive approach.  converting more of these strings to f-strings would be better but I am not sure what the code style is for this repo.

The changes outside of scripts were auto formats by black.

## Checks
I went to run the unit tests and I got this far down the dependencies before giving up.

`pip install xarray ply fastapi metpy flask dask sentry_sdk gsw pyinstrument geopy pyresample pydantic_settings babel views` 

Do you have a requirements file? 

- [ ] I ran unit tests.

### DIFF
--- a/data/mercator.py
+++ b/data/mercator.py
@@ -246,9 +246,9 @@ class Mercator(Model):
                     np.zeros(d.shape[1:]), mask=True, dtype=self.depths.dtype
                 )
 
-                depth_values[
-                    np.unravel_index(indices, depth_values.shape)
-                ] = self.depths[depths]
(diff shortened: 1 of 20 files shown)


JustinElms:
Thanks for the contribution! These scripts are pretty old and long overdue for some refactoring. 

You can find the environment requirements in `config/conda/environment.yml`. Just a note that you may also need to run `pip install visvalingamwyatt` separately since it often gets missed during the install.

JustinElms (approved):

Issue #1194, opened 2025-06-20
Add depth selector to subset panel


JustinElms:
**Priority (low, medium, high)**
Priority: medium

**Is your feature request related to a problem? Please describe.**
The subset feature currently returns a NetCDF file containing all of the depth levels available for the selected dataset. In many cases this may be more data than the user needs or create files that are too large to be processed and within the Navigator’s response time limit or crash the Navigator instance. 

**Describe the solution you'd like**
Add a dropdown or similar type menu to the Subset Panel in the area plot for 3D datasets. This menu should include an “All” option for uses who would like all of the available data. An option to select a depth range might be nice if its not too much additional work but is not required.

PR #1208, opened 2025-06-24
"Add depth slicing logic for variables

Shasan634:
#1194 
## Background
This PR adds support for depth slicing when subsetting NetCDF data. Users can now choose a specific depth range for variables that have a depth dimension.

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map/Map.jsx
+++ b/oceannavigator/frontend/src/components/Map/Map.jsx
@@ -967,7 +967,7 @@ const Map = forwardRef((props, ref) => {
     }
 
     const vectorTileGrid = new olTilegrid.createXYZ({
-      tileSize: 512,
+      tileSize: 256,
       maxZoom: MAX_ZOOM[props.mapSettings.projection],
     });
 
@@ -1045,7 +1045,7 @@ const Map = forwardRef((props, ref) => {
       mapLayers[4].setSource(null);
     } else {
       const vectorTileGrid = new olTilegrid.createXYZ({
-        tileSize: 512,
+        tileSize: 256,
         maxZoom: MAX_ZOOM[props.mapSettings.projection],
       });
 
@@ -1063,7 +1063,6 @@ const Map = forwardRef((props, ref) => {
         new VectorTile({
           format: new MVT(),
           tileGrid: vectorTileGrid,
-          tilePixelRatio: 8,
           url: `/api/v2.0/mbt/bath/{z}/{x}/{y}?projection=${props.mapSettings.projection}`,
           projection: props.mapSettings.projection,
         })
--- a/oceannavigator/frontend/src/components/Map/utils.js
+++ b/oceannavigator/frontend/src/components/Map/utils.js
@@ -86,9 +86,10 @@ export const getBasemap = (
       const shadedRelief = topoShadedRelief ? "true" : "false";
       return new TileLayer({
         preload: 1,
+        maxZoom:7-1e-10,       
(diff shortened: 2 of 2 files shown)


routes/api_v2_0.py
```diff
@@ -487,11 +487,13 @@ def subset_query(
     ),
     time: str = Query(description="", examples=["2283984000,2283984000"]),
     should_zip: str = Query("1", description="", examples=["1"]),
+    depth: str = Query(None, description="Optional depth index (e.g. 0 or 'bottom')"),
```
JustinElms, line 490:
```suggestion
    depth: str = Query(None, description="Optional depth index (e.g. 0 or 'bottom')"),
    should_zip: str = Query("1", description="", examples=["1"]),
```

This is a little nitpicky but we should group similar query components together so that users don't miss anything. I'd switch these two so that all of the data parameters (area, time, depth) are grouped. `output_format` would ideally be moved as well but this could break any existing API scripts.

routes/api_v2_0.py
```diff
@@ -513,6 +515,9 @@ def subset_query(
         timestamp=int(time_range[0]),
         endtime=int(time_range[1]),
     ) as dataset:
+        
+        if depth is not None:
+            args['depth']=depth
```
JustinElms, line 520: depth was already been added to `args` on 495 so this can be removed. A `None` value can be handled in the subset function

oceannavigator/frontend/src/components/SubsetPanel.jsx
```diff
@@ -106,15 +137,38 @@ function SubsetPanel(props) {
...
+
   const getSubsetVariables = () => {
-    setLoading(true);
-    GetVariablesPromise(props.dataset.id).then((variableResult) => {
-      setLoading(false);
+  setLoading(true);
+
+  Promise.all([
+    GetVariablesPromise(props.dataset.id),
+    GetDepthsPromise(props.dataset.id, props.dataset.variable)
```
JustinElms, line 155: Good idea to combine the promises here but since `props.dataset.variable` is inherited from the area window then the promise will never return any values if the user initially plotted a 2D variable. (Try opening an area plot with "Critical Depth" to see what I mean)

A safer approach might be to query depths when a 3d variable is selected from the variable `ComboBox` component by adding using a handler function with the onUpdate callback. The query might be cached from the users previous interactions and wouldn't be too inefficient if so. From there you can either store the depths in a state variable as you're doing or just rely on the cached data.

oceannavigator/frontend/src/components/SubsetPanel.jsx
```diff
@@ -54,6 +71,86 @@ function SubsetPanel(props) {
...
+            setOutputDepth(result.data[1].id);
+          }
+        })
+        .catch((err) => console.error("Failed to fetch depths:", err));
+    }
+  } else {
+    setAvailableDepths([]);
+    setOutputDepth(props.dataset.depth);
+  }
+};
```
JustinElms, line 152: You can condense all of this logic into a few lines by filtering for 3d variables on line 129. i.e:

```
    const variablesWithDepth = subset_variables.filter(
      (v) => selectedVariableIds.includes(v.id) && v.two_dimensional === false
    );

    if (variablesWithDepth.length > 0) {
     get depths and set state
    } else {
     set state for no depths
    }
```

JustinElms (approved):

Issue #1203, opened 2025-06-20
Expand map annotation feature capability


JustinElms:
**Priority (low, medium, high)**
Priority: medium

**Is your feature request related to a problem? Please describe.**
Per user request an option to add map annotation labels has been added to the Navigator but the labels themselves are not sufficient enough to highlight features of interest.

**Describe the solution you'd like**
Expand upon the feature to allow users to indicate which points the label corresponds to. Possible solutions include adding arrow or star icons to point to the area of interest or add drawing capabilities to the annotation layer that giver users finer control.

PR #1218, opened 2025-07-11
Improved annotation functionality

Shasan634:
#1203

## Replace vector-feature annotations with overlay manager; add directional arrows, close button, and compare-view sync
This PR replaces the old “annotations as vector features” approach with a DOM-overlay–based AnnotationOverlayManager. Users can now:

- **Add labels that point to a feature with a small arrow.

- **Rotate the arrow (down/left/up/right) by clicking the label.

- **Drag labels smoothly without accidentally rotating them.

- **Close labels via an inline × button.

## Key Features
- **Rotating arrows**: Click annotation to cycle through 4 directions (down/left/up/right)
- **Smart interactions**: Distinguishes between clicks and drags to prevent accidental rotations
- **Close button**: Inline × button for easy removal

### **AnnotationOverlayManager.jsx** (new)
- Single manager handles both maps with unified overlay lifecycle
- DOM-based overlays with CSS arrow styling in 4 directions
- Movement threshold prevents rotation during drag operations
- Compare mode attach/detach via `setSecondaryMap(map|null)`

### **Map.jsx**
- Replaced vector annotation layer with overlay manager approach
- Added `getMapCenter()` method for positioning new annotations
- Simplified annotation methods now delegate to manager
- Compare mode integration calls `setSecondaryMap()` automatically
- Updated `resetMap()` to clear overlays

### **AnnotationTextWindow.jsx**
- Enhanced with better error handling and validation
- Uses `getMapCenter()` to position annotations at map center
- Improved UX with autofocus and keyboard support

### **AnnotationButton.jsx**
- Streamlined to directly open annotation modal
- Removed complex annotation mode state management

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/Map/Map.jsx
```diff
@@ -899,26 +907,98 @@ const Map = forwardRef((props, ref) => {
...
+        modalType: "annotation",
+        showModal: true,
+      });
+    };
+    map0.on("click", onMapClick);
+    return () => {
+      map0.un("click", onMapClick);
+    };
+  }, [map0, annotationMode]);
+
```
JustinElms, line 1001: Instead of using an effect hook here to add the click listener you can just add the click listener to the map object when it gets created like we do here: https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/3833ccfbeed55a1710a4341156fa52d5d23aba4e/oceannavigator/frontend/src/components/Map/Map.jsx#L219

oceannavigator/frontend/src/components/OceanNavigator.jsx
```diff
@@ -178,6 +180,13 @@ function OceanNavigator(props) {
           return !prevCompare;
         });
         break;
+
+      case "enableAnnotationMode":
+        mapRef.current.enableAnnotationMode();
+        break;
+      case "disableAnnotationMode":
+        mapRef.current.disableAnnotationMode();
+        break;
```
JustinElms, line 189: I think you can combine these into one action since the annotation mode is disabled within the Map component once the user clicks on where they want to place the new label.

JustinElms:
I really like the arrow directionality feature you added!

A few small changes left - most are more organizational than anything:

- 'AnnotationOverlay' is handling a lot of the functionality that really belongs to the Map component. Let the map handle adding/removing overlays and treat 'AnnotationOverlay' as a component that returns jsx. You can use the Map component's 'addAnnotationLabel', 'undoAnnotationLabel', and 'clearAnnotationLabels' functions to do those things directly rather than pass the map components to the 'AnnotationOverlay' and call them in there. I think that's more consistent with React best practices but let me know if you disagree.

- You added a function to the Map that passes the central lat/lon ('centerCoord') of the current view to 'AnnotationTextWindow', then in 'AnnotationTextWindow' you pass that back to the Map component as an argument to 'mapRef.addAnnotationLabel'. Why not just get the coordinates within that function instead of pulling them out only to pass them back again? Alternatively, you could restore the click placement function that you had implemented in the previous version of this PR which was a little more user-friendly than placing the label in the centre by default.

- Move AnnotationOverlay.jsx to the map folder since its really a map component.

- The 'annotationMode' element passed to 'updateUI' isn't used.

- Can we make the label background more transparent so that users can see the underlying data?

oceannavigator/frontend/src/components/AnnotationButton.jsx
```diff
@@ -7,8 +7,12 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { withTranslation } from "react-i18next";
 
 function AnnotationButton(props) {
-  const handleAnnotation = () => {
-    props.updateUI({ modalType: "annotation", showModal: true });
+  const handleClick = () => {
+    props.updateUI({
+      modalType: "annotation",
+      showModal: true,
+      annotationMode: false,
```
JustinElms, line 14:
```suggestion
    props.updateUI({
      modalType: "annotation",
      showModal: true,
```

PR #1219, opened 2025-07-15
Functional_Component

Shasan634:
## Background
Migrating class components to functional

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/AreaWindow.jsx
+++ b/oceannavigator/frontend/src/components/AreaWindow.jsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect, useRef } from "react";
 import { Accordion, Card, Col, Row, Nav } from "react-bootstrap";
 import Button from "react-bootstrap/Button";
 import PlotImage from "./PlotImage.jsx";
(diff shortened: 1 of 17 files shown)


oceannavigator/frontend/src/components/Class4Window.jsx
```diff
@@ -1,197 +1,159 @@
...
-      size: "10x7",
-      dpi: 144,
-      models: [],
+import { useTranslation } from "react-i18next";
+
+const Class4Window = ({ dataset, class4type, plotData, init = {}, action }) => {
+  const { t: _ } = useTranslation();
+  // Track if mounted to prevent no-op errors with the Ajax callbacks.
+  const isMountedRef = useRef(false);
+
```
JustinElms, line 14: I think this was necessary for class component lifecycles but can probably be removed for functional ones.

oceannavigator/frontend/src/components/Class4Window.jsx
```diff
@@ -1,197 +1,159 @@
...
-  plotData: PropTypes.object,
+  dataset: PropTypes.string.isRequired,
+  class4type: PropTypes.string.isRequired,
+  plotData: PropTypes.object.isRequired,
   init: PropTypes.object,
   action: PropTypes.func,
 };
 
-export default withTranslation()(Class4Window);
+export default Class4Window;
```
JustinElms, line 159:
```suggestion
export default withTranslation()(Class4Window);
```

oceannavigator/frontend/src/components/ComboBox.jsx
```diff
@@ -1,347 +1,224 @@
...
-  componentWillUnmount() {
-    this._mounted = false; //Component not mounted
-  }
+  // re‑populate whenever url or incomingData change
+  useEffect(() => {
+    if (url !== lastUrl || incomingData !== undefined) {
+      setLastUrl(url);
+      populate();
+    }
+  }, [url, incomingData]);
```
JustinElms, line 40: Probably don't need `useEffect` here since any changes to data or url will force a re-render anyway (see https://react.dev/learn/you-might-not-need-an-effect - definitely worth a read). And since these ops are pretty light and it should only re-render when state/props are changed you won't need to compare to previous values. Not sure if the intent was to recreate the functionality of `componentWillUnmount` was deprecated because it led many bad design patterns and should have been removed from our code long ago.

oceannavigator/frontend/src/components/AreaWindow.jsx
```diff
@@ -12,465 +12,355 @@ import CustomPlotLabels from "./CustomPlotLabels.jsx";
...
+      "Variable",
+      "Variable Range",
+      "Colourmap",
+      "Show Bathymetry Contours",
+      "Arrows",
+      "Additional Contours",
+      "Show Selected Area(s)",
+      "Saved Image Size",
+    ].forEach((key) => t(key));
+  }, [t]);
```
JustinElms, line 123: I think this was only necessary inside the `render` function. It would be worth checking if it can be omitted once you switch to the `_()` shortcut.

JustinElms:
### Class4Window
- _Additional Models_ selector does not allow users to select additional models
- _Show Error_ dropdown not working

JustinElms (approved):

PR #1221, opened 2025-07-15
props destructuring

Shasan634:
# Destructuring DatasetSelector props and fixing defaultProps deprecation

## Summary
Refactored the DatasetSelector component to use direct prop destructuring, replaced the `__` translation alias with direct `t` usage, and converted deprecated `defaultProps` to modern default parameters to resolve React deprecation warnings.

## Problem Solved
The component was triggering a React deprecation warning:

**Warning: DatasetSelector: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.**

Screenshot 2025-08-14 132107

### 1. Updated DatasetSelector Component
- **Converted from props object to destructured parameters** in function signature
- **Replaced `defaultProps` with default parameters** to fix React deprecation warning
- **Removed** the `const __ = t;` alias assignment
- **Replaced** all instances of `__()` with `t()` throughout the component
- **Updated** all prop references from `props.propName` to direct `propName` usage

### 2. Updated Webpack Configuration
- **Removed** the DefinePlugin that was setting up global `__` and `_` aliases:
  ```javascript
  // REMOVED:
  new webpack.DefinePlugin({
    "_": "this.props.t",
    "__": "props.t",
  }),
  ```

### Before (Deprecated Pattern):
```javascript
function DatasetSelector(props) {
  const __ = t;
  // Usage: props.onUpdate, __("Dataset")
}

DatasetSelector.defaultProps = {
  showQuiverSelector: true,
  multipleVariables: false,
  // ...
};
```

### After (Modern Pattern):
```javascript
function DatasetSelector({
  onUpdate,
  showQuiverSelector = true,
  multipleVariables = false,
  // ...
  t,
}) {
  // Usage: onUpdate, t("Dataset")
}
``

## Files Changed
- `src/components/DatasetSelector.jsx` - Updated to use destructured props, default parameters, and `t` directly
- `webpack.config.js` - Removed DefinePlugin for translation aliases

## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```


oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -26,15 +26,38 @@ import "rc-slider/assets/index.css";
...
+  mountedDataset,
+  mapSettings,
+  showTimeSlider,
+  compareDatasets,
+  showCompare,
+  action,
+  t,
+}) {
+  // Using t as the translation function instead of __
+  const __ = t;
```
JustinElms, line 52: If we're including t in the component definition then it'll be more readable just to use `t` instead of `__` throughout. Just be sure to remove the plugin definition from the webpack config.

JustinElms:
Great work! Just a couple style things to fix for readability as commented above. 

The PR's documentation is lacking though. You should provide as much info for the PR template as possible so that anyone looking back will understand why we're making these changes. Including the text or just a screenshot of the error that prompted these changes would go a long way too. The PR title should also be more descriptive. You're only de-structuring props of one component but calling this PR "props destructuring" could be interpreted as something much more encompassing. Something like "Destructuring DatasetSelector props" would make it pretty clear.

oceannavigator/frontend/src/components/DatasetSelector.jsx
```diff
@@ -146,13 +161,13 @@ function DatasetSelector(props) {
...
-                  default_location: currentDataset.default_location
+                  default_location: currentDataset.default_location,
                 });
                 setDatasetVariables(variableResult.data);
                 setDatasetTimestamps(timeData);
                 setDatasetDepths(depthResult.data);
-                setOptions({
-                  ...props.options,
+                setOptionsState({
+                  ...options,
```
JustinElms, line 170:
```suggestion
                setOptionsState({
                  ...mapSettings,
```

Issue #1198, opened 2025-06-20
Update pre-defined area style to match drawn features


JustinElms:
**Priority (low, medium, high)**
Priority: low

**Is your feature request related to a problem? Please describe.**
The map interaction refactor of v9.6 allows users to display both drawn and predefined features simultaneously but the predefined area styles have not yet been updated to match the others. 

**Describe the solution you'd like**
Update the style of the predefined area features in the Map components loadFeatures method so that they have the same appearance as other areas.

PR #1222, opened 2025-07-17
Display feature labels on map and make feature styles consistent

Shasan634:
#1198
## Background
Area plotted by preset drawings had a different styling then area plotted by user.Changed preset area plot styles and also added a  new labeling button that only appears when a user draws a polygon.The  label is automatically added to the center of the drawing
## Checks
- [x] I ran unit tests.

```sh
bash run_python_tests.sh
```

### DIFF
--- a/oceannavigator/frontend/src/components/Map/utils.js
+++ b/oceannavigator/frontend/src/components/Map/utils.js
@@ -306,13 +306,13 @@ export const createMap = (
               stroke: new Stroke({
                 color: "#ffffff",
                 width: 5,
-              })
+              }),
             }),
             new Style({
               stroke: new Stroke({
                 color: "#ff0000",
                 width: 3,
-              })
+              }),
             }),
             new Style({
               geometry: new olgeom.Point(
@@ -546,7 +546,12 @@ export const createFeatureVectorLayer = (source, mapSettings) => {
                   }),
                 }),
               ];
-              const textStyle = createFeatureTextStyle(feat);
+              const textStyle = createFeatureTextStyle(
+                feat,
+                "#000",
+                "#ffffff",
+                mapSettings
+              );
               if (textStyle) styles.push(textStyle);
               return styles;
             }
@@ -565,7 +570,9 @@ export const createFeatureVectorLayer = (source, mapSettings) => {
                 }),
               }),
             ];
-            const textStyle = createFeatureTextStyle(feat);
+            const textStyle = createFeatureTextStyle(feat,                "#000",
+                "#ffffff",
+                mapSettings);
             if (textStyle) styles.push(textStyle);
             return styles;
           case "Point":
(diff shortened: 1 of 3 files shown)


JustinElms:
Labels are not drawn for selected areas:
image
This isn't related to the changes that you've made here but since this PR deals with feature styles it might be worth addressing.

oceannavigator/frontend/src/components/Map/utils.js
```diff
@@ -611,6 +633,103 @@ export const createFeatureVectorLayer = (source, mapSettings) => {
...
+                      offsetX: offsetX,
+                      offsetY: offsetY,
+                      rotation: 0,
+                    }),
+                  })
+                );
+              }
+            }
+
+            return styles;
```
JustinElms, line 732: I really like your approach here, it seems to give a pretty consistent placement for labels. A couple suggestions to help make things a little more efficient:

1. This code will be run a lot - on map interactions, panning, zooming, selecting/deselecting, changing label text, etc. If you instead move this to [updateFeatureGeometry](https://github.com/DFO-Ocean-Navigator/Ocean-Data-Map-Project/blob/4962af7ddee71496f8e6d9e370c47f155ba6df8a/oceannavigator/frontend/src/components/Map/Map.jsx#L608) and add the result as a property of the feature then you should be able to check to see if the work was already done and reuse it if so.
2. You could simplify a lot of the logic by calculating the angle between the start and end points and selecting the position based on the result. If you use the current projection coordinates for this then it should also work well for the polar projections.
3. There are a number of methods in the LineString and Geometry classes that provide some of the functions you're writing yourself here. Check out the docs to see what's available and where you could apply it. Some that come to mind are `getFirstCoordinate`, `getLastCoordinate`, `getExtent`.
4. Lastly, to keep the code clean make this a separate function.

oceannavigator/frontend/src/components/Map/utils.js
```diff
@@ -636,6 +672,57 @@ export const createFeatureVectorLayer = (source, mapSettings) => {
...
+      );
+    } else {
+      // Regular polygons
+      let centroid = feature.get("labelCentroid");
+      if (!centroid) {
+        centroid = feature.getGeometry().getInteriorPoint().getCoordinates();
+        feature.set("labelCentroid", centroid);
+      }
+      geometry = new olgeom.Point(centroid);
+    }
```
JustinElms, line 710: You could easily reduce this by adding a `centroid` property to the feature instead of `labelCentroid`.  Something like this could work:

```suggestion
  } else if (type === "Polygon") {
    if (!feature.get("centroid")) {
      let centroid = feature.getGeometry().getInteriorPoint().getCoordinates();
      feature.set("centroid", olProj.toLonLat(centroid).slice(0, 2));
    }
    geometry = new olgeom.Point(
      olProj.transform(
        feature.get("centroid"),
        "EPSG:4326",
        mapSettings.projection
      )
    );
    }
```

PR #1234, opened 2025-08-27
Dataset_Search_Feature

Shasan634:
## Checks
- [ ] I ran unit tests.

```sh
bash run_python_tests.sh
```


routes/api_v2_0.py
```diff
@@ -651,15 +580,12 @@ def filter_datasets_by_depth(
...
-                dims = db.get_all_dimensions()
-                has_depth_dims = "yes" if "depth" in dims else "no"
-                if has_depth == has_depth_dims:
-                    matching_dataset_ids.append(dataset_id)
-        except:
-            continue
+        config = DatasetConfig(dataset_id)
+        with SQLiteDatabase(config.url) as db:
+            dims = db.get_all_dimensions()
+            has_depth_dims = True if "depth" in dims else False
```
JustinElms, line 586:
``` suggestion
            has_depth_dims = "depth" in dims
```

routes/api_v2_0.py
```diff
@@ -441,6 +447,230 @@ def class4_data(
...
+
+@router.get("/datasets/filter/date")
+def filter_datasets_by_date(
+    target_date: str = Query(description="Target date in ISO format"),
+    dataset_ids: str = Query(description="Comma-separated dataset IDs to filter from"),
+):
+    """
+    Returns only matching dataset IDs for date filter.
+    """
+    dataset_id_list = [id.strip() for id in dataset_ids.split(",") if id.strip()]
```
JustinElms, line 601: If you specify the `dataset_ids` param as a list then you shouldn't have to decode the string assuming it's encoded correctly in the query.

routes/api_v2_0.py
```diff
@@ -441,6 +447,126 @@ def class4_data(
...
+                data = xr.open_mfdataset([config.url])
+            else:
+                data = xr.open_mfdataset(config.url)
+            time_vals = data.time.values
+            first_timestamp = time_vals[0].astype("datetime64[s]").astype(int)
+            last_timestamp = time_vals[-1].astype("datetime64[s]").astype(int)
+            first_time = datetime.fromtimestamp(first_timestamp, tz=timezone.utc)
+            last_time = datetime.fromtimestamp(last_timestamp, tz=timezone.utc)
+            if first_time <= parsed_date and last_time >= parsed_date:
+                matching_dataset_ids.append(dataset_id)
```
JustinElms, line 530: Rather than going through the trouble of converting the timestamps to a datetime object you can just do the inverse:
```suggestion
            if time_vals.min() <= np.datetime64(parsed_date) and time_vals.max() >= np.datetime64(parsed_date):
                matching_dataset_ids.append(dataset_id)
```

routes/api_v2_0.py
```diff
@@ -441,6 +447,126 @@ def class4_data(
...
+    """
+    dataset_id_list = [id.strip() for id in dataset_ids.split(",") if id.strip()]
+
+    point = Point([longitude, latitude])
+    shapes_dir = FilePath("/data/misc/dataset_shapes")
+
+    matching_dataset_ids = []
+
+    for dataset_id in dataset_id_list:
+        try:
```
JustinElms, line 552: No need for try/except here

